PR #207, opened 2026-04-13
fix: pass orgId through GitHub OAuth state for reliable org association

### LINKED ISSUE #204: Fix GitHub App installation org association via state parameter
## Problem
The GitHub App installation callback relies on a `current_org_id` cookie to determine which organization to associate the installation with. This is unreliable — the cookie may be stale or missing.

## Solution
Pass the `orgId` through the GitHub OAuth `state` parameter (`orgId|returnUrl` format) so the callback can reliably resolve the target organization.

## Files
- `apps/web/app/(app)/settings/integrations/github-integration-card.tsx` — encode orgId in state param
- `apps/web/app/(app)/settings/integrations/page.tsx` — pass orgId prop to card
- `apps/web/app/api/github/callback/route.ts` — extract orgId from state, use as primary org resolver
- `apps/web/app/api/github/webhook/route.ts` — ensure organizationId is set on repo upserts

redoh:
## Summary
- Encodes `orgId` in the GitHub OAuth `state` parameter (`orgId|returnUrl` format) when initiating app installation
- Extracts `orgId` from state in callback route, using it as the primary org resolver instead of relying on a cookie
- Forwards `orgId` in state when redirecting across origins
- Sets `organizationId` on repository upserts in both callback and webhook routes

Closes #204

## Files Changed
- `apps/web/app/(app)/settings/integrations/github-integration-card.tsx`
- `apps/web/app/(app)/settings/integrations/page.tsx`
- `apps/web/app/api/github/callback/route.ts`
- `apps/web/app/api/github/webhook/route.ts`

### DIFF
--- a/apps/web/app/(app)/settings/integrations/github-integration-card.tsx
+++ b/apps/web/app/(app)/settings/integrations/github-integration-card.tsx
@@ -21,9 +21,11 @@ type GitHubData = {
 export function GitHubIntegrationCard({
   data,
   appSlug,
+  orgId,
 }: {
   data: GitHubData;
   appSlug: string | null;
+  orgId: string;
 }) {
   const [isPending, startTransition] = useTransition();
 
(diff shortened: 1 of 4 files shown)


redoh:
**Review feedback addressed:**

- ✅ **🔵 Debug console.log in production** — Fixed. Simplified the verbose log that exposed cookie/orgId values.
- 👎 **🟠 orgId not verified against session** — False positive. The code already verifies membership: `organizationMember.findFirst({ userId: session.user.id, organizationId })`.
- 👎 **🟡 Pipe delimiter conflicts** — False positive. CUIDs never contain `|`, and the state is URL-encoded.
- 👎 **🟡 Manage Repos loses target="_blank"** — Intentional. Same-tab navigation is required for the redirect flow to work properly with the state parameter.
- 👎 **💡 orgId prop should accept null** — False positive. The calling page guarantees orgId exists for authenticated members.

@octopus ready for re-review.

PR #263, opened 2026-04-21
fix(billing): transaction history pagination shows correct total

### LINKED ISSUE #261: fix(billing): transaction history pagination shows wrong page count
## Problem
On `/settings/billing`, the Transaction History pagination initially shows `1 / 1` even when more transactions exist on the server. Clicking next loads the next batch and the label jumps to `2 / 2` — confusing UX.

Root cause: `totalPages` was computed from locally-loaded transactions only (first 20), not the real server-side count. Additionally, `loadMoreTransactions` returned raw Prisma objects (Decimal/Date) which Next.js refuses to serialize across the server→client boundary.

## Fix
- Fetch `totalTransactions = prisma.creditTransaction.count(...)` on the page and pass it down.
- Compute `totalPages` from the real total (fallback to filtered length when a client-side type filter is active).
- Unify the next-page button so it triggers `loadMore` transparently when the next page isn't yet loaded.
- Serialize Decimals/Dates to plain types inside the `loadMoreTransactions` server action.

## Files
- `apps/web/app/(app)/settings/billing/page.tsx`
- `apps/web/app/(app)/settings/billing/billing-settings.tsx`
- `apps/web/app/(app)/settings/billing/actions.ts`

redoh:
## Summary
- Fetch real `transactions.count` on the server and compute `totalPages` from that instead of from locally-loaded rows, so the label no longer jumps from `1/1` to `2/2` on first next.
- Unify the next-page button: it triggers `loadMore` transparently when the next page is not yet in memory.
- Serialize Decimal/Date inside `loadMoreTransactions` so the server action can return values across the server/client boundary.

Closes #261

## Test plan
- [ ] Visit `/settings/billing` with more than 20 transactions; label reads correctly (e.g. `1 / 5`).
- [ ] Next/prev buttons navigate without Next.js console errors.
- [ ] Type filter narrows pagination to filtered count.

### DIFF
--- a/apps/web/app/(app)/settings/billing/actions.ts
+++ b/apps/web/app/(app)/settings/billing/actions.ts
@@ -145,7 +145,17 @@ export async function loadMoreTransactions(
   orgId: string,
   offset: number,
   limit: number = 20,
-) {
+): Promise<
+  {
+    id: string;
+    amount: number;
+    type: string;
+    description: string | null;
+    receiptUrl: string | null;
+    balanceAfter: number;
+    createdAt: string;
+  }[]
+> {
   const session = await auth.api.getSession({ headers: await headers() });
   if (!session) return [];
 
@@ -159,10 +169,20 @@ export async function loadMoreTransactions(
 
   if (!member) return [];
 
-  return prisma.creditTransaction.findMany({
+  const rows = await prisma.creditTransaction.findMany({
     where: { organizationId: orgId },
     orderBy: { createdAt: "desc" },
     skip: offset,
     take: limit,
   });
+
+  return rows.map((t) => ({
+    id: t.id,
+    amount: Number(t.amount),
+    type: t.type,
+    description: t.description,
+    receiptUrl: t.receiptUrl,
+    balanceAfter: Number(t.balanceAfter),
+    createdAt: t.createdAt.toISOString(),
+  }));
 }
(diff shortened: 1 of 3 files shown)


redoh:
## Review response — PR #263

Thanks for the review. Status per finding:

- 🟠 **L493 — next-page blank on failed load**: Valid. `handleLoadMore` now returns a boolean; the next-page button only advances `currentPage` when the load succeeded. On server errors we keep the user on the current page instead of showing an empty table.
- 🔵 **L148 — duplicated inline return type**: Valid. Extracted `TransactionDTO` in `actions.ts` and imported it in `billing-settings.tsx`, so the server-side and client-side shapes are kept in lockstep.
- 🟡 **L336 — filter while on page > 1 shows blank**: Not reproducible. Both the `All` badge and the per-type badges already call `setCurrentPage(1)` in their `onClick`, so switching filters always snaps back to page 1. No change.
- 🟡 **L120 — totalTransactions stale after purchase**: The purchase flow redirects to Stripe Checkout and back to `/settings/billing`, which re-renders the server component and re-fetches `count`. In-page realtime updates via Pubby would be a separate feature; out of scope for this bug fix.

Pushed as `219e267`. @octopus-review PTAL.

redoh:
@octopus review feedback applied and pushed. Ready for another look.

PR #273, opened 2026-04-24
feat(github): sign install state + route through /api/github/install

### LINKED ISSUE #272: Harden GitHub App install callback with signed state + better error UX
## Background

The GitHub App install flow currently passes a plaintext `state=orgId|returnUrl` string through the OAuth-style redirect. The callback trusts whatever `orgId` shows up there, falls back to a cookie, and swallows most failures silently (redirecting to `/dashboard` with no explanation). That is fragile and reviewer-unfriendly:

- Anyone who gets a user to click a crafted URL can influence which org the installation binds to.
- If the user has somehow lost membership, is already bound to a different org, or the flow gets interrupted, we silently drop them on `/dashboard` with no feedback.
- Disconnect leaves `Repository.installationId` populated, so a later `syncRepos` can silently re-activate rows.

## Changes

- Add an HMAC-signed, expiring, single-use state (`apps/web/lib/github-install-state.ts`) — signed with a new `GITHUB_STATE_SECRET`, with Redis-backed replay protection via `jti`.
- Replace the inline GitHub install URLs that were scattered across the app (settings card, repositories page, providers banner, permission banner) with a single server-side starter at `/api/github/install?returnTo=...`. That route authenticates the user, resolves their org, signs a state, and redirects to GitHub with just `state`.
- Rework `/api/github/callback` to verify the signed state, check replay, validate membership, reject cross-org hijacking of an already-bound installation, and surface specific errors to `/settings/integrations?error=<reason>`.
- Render those errors in the settings card with a real dialog so users know what happened and how to recover.
- Make `disconnectGitHub` clear `Repository.installationId` in the same transaction that nulls the org-level id, so a later sync cannot silently re-activate rows.
- Drop the `baseUrl`/`githubAppSlug` plumbing that existed only to build install URLs on the client — callers now just use the `/api/github/install` entry point.

## Files

- `.env.example` — document `GITHUB_STATE_SECRET`
- `apps/web/app/api/github/install/route.ts` (new) — authenticated starter
- `apps/web/lib/github-install-state.ts` (new) — sign/verify/replay helpers
- `apps/web/app/api/github/callback/route.ts` — verify signed state + map errors
- `apps/web/app/(app)/settings/integrations/actions.ts` — clean up on disconnect
- `apps/web/app/(app)/settings/integrations/github-integration-card.tsx` — error dialog + confirm dialog
- `apps/web/app/(app)/settings/integrations/page.tsx` — accept allow-listed error param
- `apps/web/components/dashboard/providers-banner.tsx`, `apps/web/components/permission-banner.tsx`, `apps/web/app/(app)/repositories/repositories-content.tsx`, `apps/web/app/(app)/dashboard/page.tsx`, `apps/web/app/(app)/layout.tsx`, `apps/web/app/(app)/repositories/page.tsx` — route callers through `/api/github/install`

redoh:
## Summary

- HMAC-signed, expiring, single-use state replaces the plaintext `orgId|returnUrl` string the GitHub App install callback used to trust.
- New `/api/github/install?returnTo=...` starter authenticates the user, resolves the org server-side, and signs the state before redirecting to GitHub. Every install/manage link in the app goes through it.
- Callback now maps specific failure modes (missing/expired/malformed/replayed state, cross-org rebinding, lost membership, invalid installation id) to `?error=<reason>`, which the settings card surfaces in a dialog instead of silently dropping users on `/dashboard`.
- Disconnect clears `Repository.installationId` in the same transaction as the org-level id, so a later `syncRepos` cannot silently re-activate rows; a confirm dialog explains what happens.

## Required env

Add `GITHUB_STATE_SECRET` (>= 32 chars) to production. The helper throws at boot if it is missing or too short.

## Test plan

- [ ] Set `GITHUB_STATE_SECRET` locally, click Install GitHub App from settings, complete install, verify repos sync.
- [ ] Tamper with the `state` query param on the callback — expect `?error=invalid_state_bad_signature` and the dialog.
- [ ] Click the returned `state` link twice — second should hit `?error=replay_detected`.
- [ ] Try to bind an installation already owned by a different org — expect `?error=installation_already_bound`.
- [ ] Disconnect, then re-install with the same installation — repos come back with data intact.

Closes #272


redoh:
@octopus ready for re-review.

**Addressed:**
- 🟡 `GITHUB_STATE_SECRET` missing from env validation — added a boot-time `console.warn` in `lib/github-install-state.ts` so ops sees a misconfigured secret in logs before the first install click (the `getSecret()` throw is still the hard gate, this is just visibility).
- 🟡 Replay protection silently skipped when Redis unavailable — now logs an explicit `[github/callback] Redis unavailable — replay protection is disabled for this request` warning in the `else` branch.
- 🔵 `disconnectGitHub` action result not checked — the confirm dialog now inspects the returned `{error?}` and renders it inline; the dialog stays open on error so the user can retry.
- 💡 `ALLOWED_GITHUB_ERRORS` defined inside render — hoisted to module scope along with a derived `GitHubErrorCode` type.

**Treated as false-positive / no change:**
- 🟠 `route.ts:L68` "Installation re-bind silently skipped when org already owns installation" — this is the intended branch: if the org is already bound to the same installation id, the write is a no-op; the repo sync below still runs. No data drift, no silent regression.
- 🟡 `permission-banner.tsx:L36` "constructs install URL client-side without signing" — the component now links to `/api/github/install?returnTo=/dashboard`, which is the server-side starter that mints the signed state. No client-side URL construction remains.
- 🔵 `install/route.ts:L7` "baseUrl fallback to hardcoded localhost" — this matches the existing project-wide convention (`BETTER_AUTH_URL || "http://localhost:3000"`) and localhost is only reachable during dev.
- 💡 `safeRelativePath` fallback driven by `returnTo` — current `/settings/integrations` fallback is intentional since this starter is almost exclusively hit from that page; keeping a single, auditable default is safer than dynamic behavior for a rarely-exercised edge case.

PR #281, opened 2026-04-24
chore(infra): resolve Cloudflare real client IP in nginx

### LINKED ISSUE #280: nginx: resolve Cloudflare real client IP (CF-Connecting-IP)
## Background

Behind Cloudflare, the TCP peer nginx sees is a Cloudflare edge IP (or in Swarm mode, the ingress hop). As a result \`\$remote_addr\`, the access log, and anything derived from it show Cloudflare IPs instead of real users. This makes access logs useless for debugging, rate-limiting, and abuse triage.

## Changes

- Add \`infra/nginx/cloudflare-real-ip.conf\` with the full Cloudflare IPv4 + IPv6 allow-list plus the Docker Swarm private ranges (10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16) so the ingress hop is trusted.
- Pull the file into both nginx configs via \`include\` and wire up a custom \`log_format\` that keeps the real IP in \`\$remote_addr\` while also logging \`cf=\$http_cf_connecting_ip\` for diagnostics.

## Files

- \`infra/nginx/cloudflare-real-ip.conf\` (new)
- \`infra/nginx/nginx.conf\`
- \`infra/nginx/nginx.aws.conf\`

redoh:
## Summary

- New \`infra/nginx/cloudflare-real-ip.conf\` with the Cloudflare IPv4 + IPv6 ranges plus Docker Swarm private nets (10/8, 172.16/12, 192.168/16) so the ingress hop is trusted.
- Both nginx configs include the file and add a \`log_format\` that keeps the real IP in \`\$remote_addr\` while also logging \`cf=\$http_cf_connecting_ip\`.

## Why

Behind Cloudflare, \`\$remote_addr\` currently resolves to a Cloudflare edge IP (or, under Swarm, the ingress hop), so access logs and anything derived from it are worthless for abuse triage and rate limiting.

## Test plan

- [ ] Reload nginx on staging — \`nginx -t\` clean, service reloads without errors.
- [ ] Hit the site through Cloudflare from a known IP — access log shows the real IP as \`\$remote_addr\` plus matching \`cf=...\` field.
- [ ] Direct-hit the origin (bypassing Cloudflare from the internal network) — still logs the internal IP, no breakage.

Closes #280

### DIFF
--- a/infra/nginx/cloudflare-real-ip.conf
+++ b/infra/nginx/cloudflare-real-ip.conf
@@ -1,0 +1,36 @@
+# Cloudflare IP ranges — https://www.cloudflare.com/ips/
+# When a request comes from one of these, use CF-Connecting-IP as the real client IP.
+# $remote_addr, access log, and X-Real-IP / X-Forwarded-For all reflect the real IP after this.
+
+# Docker Swarm ingress / overlay networks — in swarm mode, the TCP peer nginx sees
+# is the ingress hop (10.0.0.x), not Cloudflare's IP. Must trust it so real_ip kicks in.
+set_real_ip_from 10.0.0.0/8;
+set_real_ip_from 172.16.0.0/12;
+set_real_ip_from 192.168.0.0/16;
+
+set_real_ip_from 173.245.48.0/20;
+set_real_ip_from 103.21.244.0/22;
+set_real_ip_from 103.22.200.0/22;
+set_real_ip_from 103.31.4.0/22;
+set_real_ip_from 141.101.64.0/18;
+set_real_ip_from 108.162.192.0/18;
+set_real_ip_from 190.93.240.0/20;
+set_real_ip_from 188.114.96.0/20;
+set_real_ip_from 197.234.240.0/22;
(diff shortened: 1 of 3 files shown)


redoh:
@octopus ready for re-review.

**Addressed:**
- 🟡 Trusting RFC-1918 ranges enables IP spoofing if nginx is ever exposed directly — added an explicit `# SECURITY:` note in `cloudflare-real-ip.conf` documenting that the RFC-1918 entries are only safe because this nginx is behind Cloudflare, with instructions to tighten to the exact Swarm overlay subnet if the deployment model ever changes.

**Treated as false-positive / no change:**
- 💡 Cloudflare IP ranges require periodic manual updates — the file already points at `https://www.cloudflare.com/ips/` on line 1 as the refresh source; no further automation is in scope for this PR.
- 💡 Log format duplication between `nginx.conf` and `nginx.aws.conf` — the two files already diverge on several other directives; extracting the `log_format` to a third include would add a new moving part without reducing net maintenance burden. Will revisit if the duplication grows.

PR #283, opened 2026-04-24
fix(embeddings): batch by token cap to stay under OpenAI 300k limit

### LINKED ISSUE #282: Embeddings: batch by token cap to avoid OpenAI 300k-token request limit
## Background

`createEmbeddings` in `apps/web/lib/embeddings.ts` currently batches inputs by item count (512 per request). OpenAI's embeddings endpoint also enforces a **300,000 token** total-request cap. For repos with many large files, a 512-item batch can comfortably blow past that cap and fail the whole request, wasting a full retry on a single bad batch.

## Changes

- Pre-truncate each input to the existing `MAX_EMBEDDING_CHARS` limit once (not inside the batch loop).
- Switch the batcher to walk the input list and close a batch when it either hits 512 items OR approaches 250k estimated tokens (chars/3, headroom against the 300k cap).
- Small helper `estimateTokens` and named constants so the limits are documented.

## Files

- `apps/web/lib/embeddings.ts`

redoh:
## Summary

- Walk inputs and close a batch at `min(512 items, ~250k tokens)` using a chars/3 upper-bound token estimate, instead of blindly batching 512 at a time.
- Pre-truncate each input to the per-item `MAX_EMBEDDING_CHARS` cap once (outside the loop).
- Name the constants (`MAX_BATCH_TOKENS`, `MAX_BATCH_ITEMS`, `CHARS_PER_TOKEN_ESTIMATE`) so the 300k ceiling is documented.

## Why

OpenAI's embeddings endpoint enforces a 300,000-token cap per request. For repos with many large files, the old 512-item batcher could easily exceed that ceiling and fail the whole request — wasting the retry and forcing us to re-embed everything.

## Test plan

- [ ] Index a repo with > 500 files — all batches succeed, no 300k-token error in logs.
- [ ] Index a repo where a single file is near `MAX_EMBEDDING_CHARS` — it's truncated, not rejected.
- [ ] Index a tiny repo — still fits in one batch, no regression.

Closes #282

### DIFF
--- a/apps/web/lib/embeddings.ts
+++ b/apps/web/lib/embeddings.ts
@@ -13,9 +13,19 @@ function getClient(): OpenAI {
   return openai;
 }
 
-// text-embedding-3-large max: 8191 tokens
+// text-embedding-3-large max: 8191 tokens per input
 // Conservative limit: ~3 chars/token for code → 24000 chars stays safely under 8191 tokens
 const MAX_EMBEDDING_CHARS = 24_000;
+
+// OpenAI enforces a 300,000 token total-request cap for embeddings.
+// Keep headroom: target ~250k and estimate tokens as chars/3 (worst case for code).
+const MAX_BATCH_TOKENS = 250_000;
+const MAX_BATCH_ITEMS = 512;
+const CHARS_PER_TOKEN_ESTIMATE = 3;
+
+function estimateTokens(text: string): number {
+  return Math.ceil(text.length / CHARS_PER_TOKEN_ESTIMATE);
+}
 
 export async function createEmbeddings(
   texts: string[],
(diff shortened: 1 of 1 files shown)


redoh:
@octopus these findings look like false-positives — please take another look.

- 🟡 "Oversized single item silently creates over-budget batch" — every input is pre-truncated to `MAX_EMBEDDING_CHARS` (24,000 chars) before entering the batch loop. At the worst-case chars/token ratio of 3, a single item tops out at ~8,000 tokens — well under `MAX_BATCH_TOKENS` (250,000). A single item physically cannot exceed the batch budget, so the `batchEnd > batchStart` guard never produces an over-budget batch.
- 💡 "`estimateTokens` should use truncated length, not original" — it already does. The batch loop iterates over the `truncated` array (`truncated[batchEnd]`), not `validTexts`.
- 💡 "No test coverage for dynamic batching" — agreed this would be nice, but adding a test file is out of scope for this fix; filing it as a follow-up.

No code changes on this PR.

PR #294, opened 2026-04-28
feat(cancel): redis pub/sub for cross-process abort

### LINKED ISSUE #293: Cancel signals don't reach worker processes (in-memory abort registries)
## Problem

`abortIndexing(repoId)` and `abortAnalysis(repoId)` consult an in-memory `Map<string, AbortController>` and only abort if the controller lives in the same process. In production (multi-replica web + worker), the abort request lands on a process that has no controller for that repo, so the long-running job keeps running.

## Fix

Add a Redis pub/sub bus (`lib/cancel-bus.ts`) that fans out cancel requests across processes:
- `publishCancel(kind, repoId)` publishes on `octopus:cancel`.
- `onCancel(kind, handler)` subscribes once per process; both abort modules register their own handler that aborts the local controller if present.

Local behavior is preserved: the abort function still aborts in-process synchronously and also publishes for any other process holding the controller.

## Files
- `apps/web/lib/cancel-bus.ts` (new)
- `apps/web/lib/analysis-abort.ts`
- `apps/web/lib/indexing-abort.ts`

redoh:
## Summary
- New `lib/cancel-bus.ts` — Redis pub/sub on `octopus:cancel`, lazily connects only when a publisher or subscriber is needed; falls back to no-op if `REDIS_URL` is not set.
- `analysis-abort.ts` and `indexing-abort.ts` register a one-time subscriber and publish on every abort, so any process holding the controller will abort.
- Local in-process abort still happens synchronously — the bus is purely additive.

Closes #293

## Test plan
- [ ] In dev with two `next dev` instances pointed at the same Redis: start an indexing job on instance A, hit the abort endpoint on instance B → job stops.
- [ ] With `REDIS_URL` unset: no errors logged, local-only behavior intact.
- [ ] Repeated abort calls don't leak listeners (handler set stays bounded).

### DIFF
--- a/apps/web/lib/analysis-abort.ts
+++ b/apps/web/lib/analysis-abort.ts
@@ -1,8 +1,21 @@
-// In-memory registry of active analysis AbortControllers.
-// Keyed by repoId so we can cancel a specific repo's analysis.
+import { onCancel, publishCancel } from "./cancel-bus";
+
 const controllers = new Map<string, AbortController>();
 
+let listenerRegistered = false;
+function ensureListener() {
+  if (listenerRegistered) return;
+  listenerRegistered = true;
+  onCancel("analysis", (repoId) => {
+    const c = controllers.get(repoId);
+    if (!c) return;
+    c.abort();
+    controllers.delete(repoId);
+  });
+}
+
 export function createAnalysisAbortController(repoId: string): AbortController {
+  ensureListener();
   const existing = controllers.get(repoId);
   if (existing) existing.abort();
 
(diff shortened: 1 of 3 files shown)


redoh:
@octopus updates pushed.

- 🟡 **subscribed-flag race** — fixed in 76146fb. Added a `subscribing` in-flight guard, only flip `subscribed` on the resolve path, and disconnect+null the subscriber on reject so the next call can retry.
- 🟡 **subscriber `maxRetriesPerRequest: null`** — keeping as is. ioredis requires `null` for subscriber-mode connections to allow blocking commands; a finite limit would break subscribe semantics. The subscriber retries via reconnection backoff, not blocking the event loop.

PR #296, opened 2026-04-28
feat(reliability): boot-time reconcile of stale repo states

### LINKED ISSUE #295: Repos stay stuck in 'indexing' / 'analyzing' after a process crash
## Problem

If the Next.js server is killed mid-indexing or mid-analysis, `Repository.indexStatus = 'indexing'` and `analysisStatus = 'analyzing'` rows are left as-is on disk. The next process boot has no idea those jobs are dead, and the UI shows them as in-progress forever.

## Fix

Add `lib/boot-reconciler.ts` that runs on `register()` and resets any rows that have been in those states for more than 10 minutes:
- `indexStatus = 'indexing' && updatedAt < now - 10m` → `'pending'`
- `analysisStatus = 'analyzing' && updatedAt < now - 10m` → `'none'`

The 10-minute window is comfortable enough not to clobber genuinely live jobs (a healthy run touches `updatedAt` regularly via progress updates) and short enough to recover quickly after a crash.

## Files
- `apps/web/lib/boot-reconciler.ts` (new)
- `apps/web/instrumentation.ts`

redoh:
## Summary
- `lib/boot-reconciler.ts`: on server boot, reset `indexStatus = 'indexing'` rows older than 10 min to `'pending'`, and `analysisStatus = 'analyzing'` rows older than 10 min to `'none'`.
- Wired into `instrumentation.register()` so it runs once before queue workers start.
- Logs the count when anything is reset; swallows errors so a DB blip doesn't prevent startup.

Closes #295

## Test plan
- [ ] Start a real indexing job, kill the process, restart — within 10 minutes UI still shows 'indexing'; after 10 minutes (or on next restart with `updatedAt` older than cutoff) the row resets to 'pending'.
- [ ] In-progress jobs (recent `updatedAt`) are untouched on boot.
- [ ] Confirm log line `[boot-reconciler] reset stale rows: indexing=N analyzing=M` when applicable.

### DIFF
--- a/apps/web/instrumentation.ts
+++ b/apps/web/instrumentation.ts
@@ -1,6 +1,9 @@
 export async function register() {
   // Only start queue workers on the server (not during build or edge runtime)
   if (process.env.NEXT_RUNTIME === "nodejs") {
+    const { reconcileStaleRepoStates } = await import("./lib/boot-reconciler");
+    await reconcileStaleRepoStates();
+
     const { startQueue } = await import("./lib/queue");
     const boss = await startQueue();
 
--- a/apps/web/lib/boot-reconciler.ts
+++ b/apps/web/lib/boot-reconciler.ts
@@ -19,6 +19,8 @@ export async function reconcileStaleRepoStates() {
       console.log(
         `[boot-reconciler] reset stale rows: indexing=${indexing.count} analyzing=${analyzing.count}`,
       );
+    } else {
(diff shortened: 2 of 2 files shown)


redoh:
@octopus updates pushed.

- 💡 **silent on clean boot** — fixed in e966bd9. Now always logs (`reset stale rows: ...` or `no stale rows to reset`) so the reconciler is observable in logs even on the common no-op path.

PR #298, opened 2026-04-28
fix(embeddings): split batch on 300k-token 400, tighter estimate

### LINKED ISSUE #297: Embedding batches occasionally exceed OpenAI's 300k-token request cap
## Problem

`createEmbeddings()` estimates tokens as `chars / 3` and targets a 250k-token batch. For dense content (lock files, `.d.ts`, hex blobs, CJK source), real tokens-per-char can be ~2 (cl100k/o200k often hits 1+ tokens per CJK char). When the estimate underpredicts, the batch silently exceeds OpenAI's 300k-token-per-request cap and the whole repo index fails on a single 400.

## Fix

Two changes in `lib/embeddings.ts`:

1. **Better estimate.** ASCII counts as `chars/2`; non-ASCII counts as 1 token per char. Bring the target down to 200k for headroom.
2. **Recursive split fallback.** If a batch still trips the 300k cap (matches `/maximum request size|tokens per request/i` on a 400), bisect and retry. Single-item batches still throw — at that point the truncation step earlier in the function already constrains us.

## Files
- `apps/web/lib/embeddings.ts`

redoh:
## Summary
- Token estimate now counts ASCII as `chars/2` and non-ASCII as 1 token/char (closer to cl100k/o200k worst case for CJK and dense code).
- Per-request target dropped from 250k to 200k for headroom.
- New `embedBatch()` helper recursively bisects the batch when OpenAI returns a 300k-cap 400 (`maximum request size|tokens per request`), so a single dense batch no longer tanks the whole repo index.

Closes #297

## Test plan
- [ ] Index a repo with a large lockfile or generated `.d.ts` — completes without 400.
- [ ] Index a repo with CJK comments/strings — completes without 400.
- [ ] Force a small `MAX_BATCH_TOKENS` locally and confirm the bisect path is exercised.
- [ ] Non-token-cap 400s (auth, etc.) still throw — no silent retries.

### DIFF
--- a/apps/web/lib/embeddings.ts
+++ b/apps/web/lib/embeddings.ts
@@ -18,13 +18,20 @@ function getClient(): OpenAI {
 const MAX_EMBEDDING_CHARS = 24_000;
 
 // OpenAI enforces a 300,000 token total-request cap for embeddings.
-// Keep headroom: target ~250k and estimate tokens as chars/3 (worst case for code).
-const MAX_BATCH_TOKENS = 250_000;
+// Keep headroom: target ~200k. Dense content (lock files, .dts, hex blobs, CJK)
+// can tokenize at ~2 chars/token, so ASCII gets chars/2 and non-ASCII counts as
+// 1 token per char (CJK in cl100k/o200k often hits 1+ tokens per char).
+const MAX_BATCH_TOKENS = 200_000;
 const MAX_BATCH_ITEMS = 512;
-const CHARS_PER_TOKEN_ESTIMATE = 3;
 
 function estimateTokens(text: string): number {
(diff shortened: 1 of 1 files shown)


redoh:
@octopus updates pushed.

- 🟡 **surrogate pairs in `estimateTokens`** — fixed in 119d7a6. Walk the string by code point (`codePointAt` + step by 2 for surrogate pairs) instead of UTF-16 code units, so emoji and supplementary-plane CJK are counted once.
- 💡 **regex error-message match may miss future OpenAI wording** — keeping as is. The fallback path is intentional best-effort: a future wording change would mean we throw on that batch (current behavior before this PR), not crash the worker. The bisect was added specifically for the well-documented 300k-token cap, and rephrasing the regex to be looser would risk swallowing unrelated 400s.

PR #300, opened 2026-04-28
fix(qdrant): deterministic UUID mapping for non-UUID point IDs

### LINKED ISSUE #299: Qdrant rejects feedback-pattern upserts when point ID is a CUID
## Problem

Qdrant requires point IDs to be `uint64` or UUID. We pass in Prisma CUIDs (`clxxx...`) directly, which Qdrant rejects, so feedback-pattern upserts fail.

## Fix

In `lib/qdrant.ts`, hash any non-UUID string into a deterministic UUIDv5-shaped value (sha1, set version+variant nibbles). Store the original ID in `payload.issueId` so we can still recover it on read. Re-upserts of the same logical ID stay idempotent because the mapping is deterministic.

## Files
- `apps/web/lib/qdrant.ts`

redoh:
## Summary
- `toQdrantId()` returns the input unchanged if it already matches the UUID shape; otherwise sha1-hashes it into a UUIDv5-shaped value (correct version + variant nibbles).
- Original ID preserved in `payload.issueId` for read paths.
- Both upsert paths (with-sparse and dense fallback) use the mapped ID.

Closes #299

## Test plan
- [ ] Trigger a feedback-pattern upsert with a Prisma CUID — succeeds; previously 4xx'd.
- [ ] Re-upsert with the same CUID — overwrites the same point (verified via `qdrant.retrieve`).
- [ ] Pre-existing UUID IDs continue to round-trip unchanged.

### DIFF
--- a/apps/web/lib/qdrant.ts
+++ b/apps/web/lib/qdrant.ts
@@ -2,8 +2,11 @@ import { createHash } from "crypto";
 import { QdrantClient } from "@qdrant/js-client-rest";
 import { generateSparseVector } from "@/lib/sparse-vector";
 
-// Qdrant point IDs must be uint or UUID. Map non-UUID strings (e.g. Prisma CUIDs)
-// deterministically into a UUIDv5-shaped string so re-upserts stay idempotent.
+// Qdrant point IDs must be uint or UUID. Existing UUID inputs pass through
+// unchanged (so callers using crypto.randomUUID() are unaffected); non-UUID
+// strings (e.g. Prisma CUIDs) are mapped deterministically into a UUIDv5-shaped
+// string. Applied at every upsert call site so the helper is the single source
+// of truth and any future caller passing a non-UUID is handled correctly.
 function toQdrantId(id: string): string {
   if (/^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i.test(id)) {
     return id;
@@ -76,7 +79,7 @@ export async function upsertChunks(
   // Qdrant accepts max 100 points per request
   for (let i = 0; i < points.length; i += 100) {
     const batch = points.slice(i, i + 100).map((p) => ({
-      id: p.id,
+      id: toQdrantId(p.id),
       vector: p.sparseVector
         ? { "": p.vector, [SPARSE_VECTOR_NAME]: p.sparseVector }
         : p.vector,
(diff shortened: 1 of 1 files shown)


redoh:
@octopus updates pushed (0b66353).

Addressed the **partial-rollout / inconsistency** concern: `toQdrantId` is now applied at every upsert call site (`upsertChunks`, `upsertKnowledgeChunks`, `upsertReviewChunks`, `upsertChatChunk`, `upsertDiagramChunk`, `upsertDocsChunks`, `upsertFeedbackPattern`).

Why this is safe and not a breaking change for existing data:
- All other upsert callers pass `crypto.randomUUID()` (verified at the chat/diagram/code-chunk call sites). The early-return regex passthrough means valid UUIDs map to themselves, so existing Qdrant point IDs in those collections are unchanged. No re-index required.
- Only the feedback-patterns collection actually had non-UUID inputs (Prisma `issue.id` CUIDs), and that collection was already failing upserts before this PR — so its breaking change is "0 stored points → working stored points," not "old IDs invalidated."

On the **brittle UUID construction** point: keeping the hand-rolled construction. The repo doesn't depend on a `uuid` package, and the v5-shape derivation is small (3 lines) and well-commented. Pulling in a dep for one helper isn't worth it. The construction itself is correct (verified version+variant nibble placement at hex offsets 12 and 16).

redoh:
@octopus updates pushed (1762bc0).

**Consistency — `originalId` now preserved in all collections.** Added `applyQdrantId(id, payload)` that resolves the Qdrant id and, when transformation actually happens, stamps `payload.originalId` so the caller's id is recoverable on read. Applied at every upsert call site (chunks, knowledge, review, chat, diagram, docs, feedback). UUID-keyed callers are pure passthrough — no payload bloat for them.

Feedback keeps the explicit `issueId` key on top of `originalId` since `issueId` is the semantic name callers might expect for that collection.

**On the breaking-change flag:** I want to push back gently here. There are no orphaned non-UUID point IDs in Qdrant right now. The non-feedback collections all generate ids via `crypto.randomUUID()` (verified at every call site) — those stay as-is via the regex passthrough, so nothing in chat/diagram/code/knowledge/review/docs is invalidated. The only collection that ever received a non-UUID was `feedback_patterns`, where upserts were rejected by Qdrant before this PR — i.e. there were no valid points there to orphan. So there is no migration to run and no data to back-fill.

`tsc --noEmit` clean.

PR #306, opened 2026-04-28
feat(usage): activity-based usage page

### LINKED ISSUE #305: Redesign /usage around user-facing activities, not raw operation names
## Why

The current /usage page shows raw `operation` codes (`review-validation`, `chat-rerank`, `package-analyze-deep-dive`, etc.) and groups by model only. None of those names mean anything to a user looking at their bill — they want to know how many reviews, indexings, and chat messages they ran, and how that maps to cost.

## Plan

Group operations into three user-facing activities:
- **Code Reviews** — review/validation/rerank/finding verification, plus issue-content generation and feedback classification.
- **Repo Indexing** — embeddings, analyze, repo/daily summaries, knowledge enhancement, deep-dive package analysis.
- **Chat & Assistant** — chat, slack-command, related rerank/embedding/title.

Per-activity stats: total cost, total internal calls, and "unit calls" (e.g. number of `review` calls = number of reviews actually run, not the auxiliary verification calls). Per-model rows under each activity for breakdown.

Daily trend stays but switches from `(input, output, calls)` to per-day, per-model rollup so we can plot cost-per-day correctly.

## Files
- `apps/web/app/(app)/usage/page.tsx`

redoh:
## Summary
- Group raw operations into three user-facing activities (Code Reviews / Repo Indexing / Chat & Assistant) with curated lists and \"unit operations\" so we can show 'reviews run' separate from 'verification calls made'.
- Per-activity card shows cost + internal calls + unit calls and a friendlier model label (Claude Opus 4.6, Gemini 2.5 Pro, etc.).
- Replace raw `byModel`/`byOperation` queries with a single `byModelOperation` group; daily trend now rolls up per (day, model) so we can plot cost over time correctly.
- Drops unused imports + the `IconBrain`/`IconApi`/`IconChartBar` references.

Closes #305

## Test plan
- [ ] Visit `/usage` for an org with mixed activity — three activity cards render with correct cost split.
- [ ] Month switcher still works.
- [ ] Daily trend chart renders without empty rows.
- [ ] Spot-check totals add up vs. the previous \"by model\" view.

### DIFF
--- a/apps/web/app/(app)/usage/page.tsx
+++ b/apps/web/app/(app)/usage/page.tsx
@@ -9,18 +9,102 @@ import {
   CardTitle,
 } from "@/components/ui/card";
 import {
-  IconBrain,
-  IconApi,
-  IconChartBar,
   IconCurrencyDollar,
   IconChevronLeft,
   IconChevronRight,
   IconWallet,
   IconArrowRight,
+  IconGitPullRequest,
+  IconReceipt,
 } from "@tabler/icons-react";
-import { getModelPricing, calcCost, formatUsd, formatNumber } from "@/lib/cost";
+import { getModelPricing, calcCost, formatUsd } from "@/lib/cost";
 import { getOrgBalance } from "@/lib/credits";
 import Link from "next/link";
+
+// ── Activity categories ──────────────────────────────────────────────
+// Group internal pipeline operations into user-facing activities.
+
+type Activity = {
+  key: string;
+  label: string;
+  description: string;
+  operations: string[];
+  // Operations whose call count maps 1:1 to a user-visible unit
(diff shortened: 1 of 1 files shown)


redoh:
@octopus reviewed.

- 🟡 **`shortModelLabel` ordering** — keeping as is. The finding's own description confirms `claude-opus-4-6` and `claude-sonnet-4-6` already come before the shorter `claude-opus-4`/`claude-sonnet-4` checks, so matching is correct. No change needed.
- 💡 **non-null assertion on `activityStats.get()`** — keeping as is. `activityStats` is seeded with one entry per `ACTIVITIES` member immediately above the loop, and the loop only looks up keys belonging to that same constant array, so the assertion is sound.

PR #308, opened 2026-04-28
fix(sidebar): scrollable sidebar with safe-area inset

### LINKED ISSUE #307: Sidebar bottom nav clipped on small screens; ignores iOS safe area
## Problem

When the sidebar is taller than the viewport (long org list, mobile), the bottom nav and account row are clipped instead of scrolling. On iOS the bottom border also runs into the home indicator.

## Fix

In `components/app-sidebar.tsx`:
- Make the sidebar root `overflow-y-auto`.
- Drop `flex-1` from the nav so it doesn't push the bottom group off-screen; instead, the bottom nav uses `mt-auto` to stay glued to the bottom when there is room.
- Add `pb-[max(0.75rem,env(safe-area-inset-bottom))]` on the account row so it clears the iOS home indicator.

## Files
- `apps/web/components/app-sidebar.tsx`

redoh:
## Summary
- Sidebar root is `overflow-y-auto` so a tall org list/nav can scroll instead of clipping.
- Bottom nav uses `mt-auto` instead of `flex-1` on the main nav so it stays at the bottom when there is room and doesn't get pushed off when there isn't.
- Account row applies `pb-[max(0.75rem,env(safe-area-inset-bottom))]` to clear the iOS home indicator.

Closes #307

## Test plan
- [ ] Resize the window vertically: sidebar scrolls smoothly; bottom nav stays accessible.
- [ ] On iOS Safari (or DevTools device emulation), the account row sits above the home indicator.
- [ ] Collapsed sidebar still renders the bottom group correctly.

### DIFF
--- a/apps/web/components/app-sidebar.tsx
+++ b/apps/web/components/app-sidebar.tsx
@@ -295,7 +295,7 @@ function SidebarContent({
   const [couponOpen, setCouponOpen] = useState(false);
 
   return (
-    <div className="flex h-full flex-col bg-sidebar text-sidebar-foreground">
+    <div className="flex h-full flex-col overflow-y-auto bg-sidebar text-sidebar-foreground">
       {collapsed ? (
         <div className="flex flex-col items-center gap-2 border-b py-3">
           {onToggleCollapse && (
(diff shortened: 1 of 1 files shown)


redoh:
@octopus reviewed.

- 🟡 **`overflow-y-auto` may clip sticky/fixed children and tooltips** — keeping as is. The tooltip primitives in this sidebar render via Radix portals (rendered into `document.body`, not nested inside the scroll container), and the command palette is also a portal. The bottom row is sticky-via-`mt-auto`, not `position: sticky`, so it doesn't depend on the scroll container's stacking context. Visual QA on long sidebars + iOS Safari shows tooltips/command palette displaying correctly above the sidebar bounds.

PR #323, opened 2026-04-30
feat(knowledge): pin documents to always include in reviews

### LINKED ISSUE #317: Knowledge Center rules applied inconsistently to reviews
## Background

External feedback (Tsinghua University researcher using Octopus on `rcore-os/tgoskits`):

> I tried adding rules through the "Knowledge Center", but they seem to apply inconsistently (sometimes respected, sometimes ignored). More granular control over the review prompt would be ideal.

## Diagnosis

Knowledge chunks are retrieved by embedding similarity + Cohere rerank in `apps/web/lib/reviewer.ts:1164-1184`:

```ts
searchKnowledgeChunks(org.id, queryVector, 25, rerankQuery)
// ...
rerankDocuments(rerankQuery, rawKnowledgeChunks, {
  topK: 8,
  scoreThreshold: 0.20,
  minResults: 1,
  ...
})
```

If the PR diff is semantically far from the rule's wording (e.g. a generic naming-convention rule vs. a refactor diff), the rule never makes it into the prompt. From the user's perspective the rule is "ignored."

## Proposed fix

Add an `alwaysInclude: boolean` flag per knowledge entry. At review time, fetch always-include chunks separately from the similarity search and union them into `knowledgeContext` (with dedup).

### Scope
- `packages/db/prisma/schema.prisma` — migration: `alwaysInclude Boolean @default(false)` on the knowledge entry model
- `apps/web/lib/knowledge*.ts` — `getAlwaysIncludeKnowledge(orgId)` helper
- `apps/web/lib/reviewer.ts:1164-1195` — union with similarity results, dedup by id
- `apps/web/app/(app)/knowledge/` — toggle per entry in UI

### Acceptance criteria
- Toggle on → rule appears in `{{KNOWLEDGE_CONTEXT}}` of every review for that org, regardless of diff
- Toggle off → existing similarity-based behavior
- No duplicate inclusion if a rule is also returned by similarity search

redoh:
Closes #317

## Summary
- Adds `alwaysInclude` flag on `KnowledgeDocument`. Pinned docs bypass embedding-similarity retrieval and are concatenated into the review's knowledge context every time.
- Per-doc cap 8KB, total cap 24KB, with truncation marker. Dedup against similarity results by title.
- Surfaced as a Pin switch in the Knowledge Center; status `ready` required (so users do not pin docs that are still indexing).
- Recorded in `KnowledgeAuditLog`.

## Why
External feedback (Tsinghua researcher) reported Knowledge Center rules being applied inconsistently. Root cause: rules whose wording is semantically distant from the diff are dropped by the similarity search. Pinning gives users a guaranteed channel for must-have rules without removing the similarity layer.

## Migration
A manual migration is included locally at `packages/db/prisma/migrations/20260430102407_add_knowledge_always_include/migration.sql` (the migrations dir is gitignored — apply via the standard manual deploy flow):

```sql
ALTER TABLE "public"."knowledge_documents" ADD COLUMN "alwaysInclude" BOOLEAN NOT NULL DEFAULT false;
CREATE INDEX "knowledge_documents_organizationId_alwaysInclude_deletedAt_idx"
  ON "public"."knowledge_documents"("organizationId", "alwaysInclude", "deletedAt");
```

## Test plan
- [x] `bun run --cwd apps/web lint` — no new warnings
- [x] `bun run --cwd apps/web typecheck` — clean
- [x] `bun run --cwd apps/web test` — 245 pass / 0 fail (4 new tests in `knowledge-context.test.ts`)
- [ ] Apply migration in dev DB
- [ ] Verify pinned doc appears in `{{KNOWLEDGE_CONTEXT}}` of every review (check log line: `N pinned + M/K from search`)
- [ ] Verify unpinned doc still uses similarity-based retrieval
- [ ] Verify Pin toggle disabled while doc is `processing`/`error`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- Knowledge Center: pin documents to always include in every review, regardless of diff similarity (#317)
+
 ## [1.0.14] - 2026-04-29
 
 ### Added
--- a/apps/web/app/(app)/knowledge/actions.ts
+++ b/apps/web/app/(app)/knowledge/actions.ts
@@ -573,6 +573,60 @@ Output ONLY the enhanced document content. Do not add meta-commentary or explana
   }
 }
 
+export async function setKnowledgeAlwaysInclude(
+  documentId: string,
+  alwaysInclude: boolean,
+): Promise<{ error?: string }> {
+  const session = await auth.api.getSession({
+    headers: await headers(),
+  });
+  if (!session) redirect("/login");
+
+  const cookieStore = await cookies();
+  const orgId = cookieStore.get("current_org_id")?.value;
+  if (!orgId) return { error: "No organization selected." };
+
+  const doc = await prisma.knowledgeDocument.findUnique({
+    where: { id: documentId },
+    select: {
+      organizationId: true,
+      organization: {
+        select: {
+          members: {
+            where: { userId: session.user.id, deletedAt: null },
+            select: { id: true },
+          },
+        },
+      },
+    },
+  });
+
+  if (!doc || doc.organization.members.length === 0) {
+    return { error: "Document not found." };
+  }
+  if (doc.organizationId !== orgId) {
+    return { error: "Document does not belong to this organization." };
+  }
+
+  await prisma.knowledgeDocument.update({
+    where: { id: documentId },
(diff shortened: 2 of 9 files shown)


redoh:
@octopus-review thanks for the review.

**False positive — reaction 👎.**

The 💡 finding "Pin toggle accessible on non-ready docs via keyboard even when visually disabled" is incorrect. Radix UI's `Switch.Root` with `disabled={true}` sets the underlying button's `disabled` HTML attribute and `aria-disabled`, which blocks **both** mouse clicks **and** keyboard activation (Space / Enter). Clicks on the wrapping `<label>` would normally forward focus to the inner control, but since that control is disabled, the activation never fires.

So `onCheckedChange` cannot run while `doc.status !== "ready"` — the `disabled` prop is the standard radix pattern and is sufficient. No code change.

PR #343, opened 2026-05-01
fix(cost): skip credit check for community orgs

### LINKED ISSUE #340: fix(cost): skip credit check for community orgs
## Problem

Community orgs are created via `getOrCreateCommunityOrg` with `type: 2` and `freeCreditBalance: 0`. This causes `getOrgSpendLimitStatus` to return `{ blocked: true, reason: "no_credits" }` because `totalCredits <= 0`.

As a result, `createEmbeddings` (which calls `isOrgOverSpendLimit`) silently returns empty vectors for every community review, breaking the github-action review flow for free / public-repo users.

## Fix

In `apps/web/lib/cost.ts`, add an early bypass for `type === 2`:

```ts
if (org.type === 2) return { blocked: false };
```

Community orgs are already rate-limited via `communityDailyReviewLimit` (default 5/day) at the route layer, so the credit check is redundant for them.

redoh:
## Summary
- Community orgs (`type === 2`) bypass the credit/spend-limit gate in `getOrgSpendLimitStatus` because they're rate-limited by `communityDailyReviewLimit` at the route layer.
- Without this, `createEmbeddings` silently returned `[]` for every community review since their `freeCreditBalance` is 0, breaking the public-repo github-action flow.

## Test plan
- [ ] Trigger a github-action review without an `OCTOPUS_API_KEY` (community mode) and confirm embeddings/qdrant calls succeed
- [ ] Confirm paid orgs with no credits still get blocked

Closes #340

### DIFF
--- a/apps/web/lib/cost.ts
+++ b/apps/web/lib/cost.ts
@@ -1,4 +1,5 @@
 import { prisma } from "@octopus/db";
+import { ORG_TYPE } from "@/lib/org-types";
 
 type ModelPricing = { input: number; output: number };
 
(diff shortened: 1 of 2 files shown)


redoh:
Addressed the magic-literal feedback in 5e8223a:

- Added `apps/web/lib/org-types.ts` exporting `ORG_TYPE = { STANDARD: 1, COMMUNITY: 2, FRIENDLY: 3 } as const`
- Replaced `org.type === 2` with `org.type === ORG_TYPE.COMMUNITY` in `cost.ts`

Kept the change minimal — `route.ts:44` still has the same literal but it's already merged via #344, will adopt the constant in a follow-up rather than expanding this PR's scope.

@octopus updates ready for review.

PR #367, opened 2026-05-11
feat(repositories): show resolved model in repo AI Models dropdowns

redoh:
## Summary
- Repository detail page's "AI Models" section now surfaces the resolved fallback in the dropdowns: `(Org Default — <model name>)` instead of a generic `(Org Default)`.
- Resolution order: repo override → org `defaultModelId` → `AvailableModel.isPlatformDefault` → hardcoded `claude-sonnet-4-6` / `text-embedding-3-large` (imported from `lib/ai-client`, matching the settings page).
- Same treatment for both Review & Chat Model and Embedding Model dropdowns.

## Test plan
- [ ] Visit `/repositories?repo=<id>` for a repo that has no per-repo override → dropdowns should show "(Org Default — <name>)" matching whatever the org's defaultModelId resolves to.
- [ ] Set a repo override → dropdown shows the selected model.
- [ ] If the org has no defaultModelId, label should reflect the platform default (or the hardcoded fallback).

### DIFF
--- a/apps/web/app/(app)/repositories/page.tsx
+++ b/apps/web/app/(app)/repositories/page.tsx
@@ -94,7 +94,7 @@ export default async function RepositoriesPage({
   } as const;
 
   // Step 2: All queries in parallel
-  const [repos, totalCount, allRepoNames, favoriteRepos, otherOrgMemberships, availableModels, bitbucketIntegration] = await Promise.all([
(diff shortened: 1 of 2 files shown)


redoh:
@octopus updates pushed.

Addressed in 123c0cb:
- 🟡 `platformDefaults` query was sequential — moved into the existing `Promise.all([...])` block alongside the other 7 queries. The sequential `await` after destructuring is gone; now adds zero extra round-trips.

Ready for re-review.

PR #381, opened 2026-05-18
fix(email): claim credit-low cooldown atomically to prevent duplicates

### LINKED ISSUE #378: fix: prevent duplicate credit-low emails under concurrent events
## Summary

The 24h credit-low email cooldown was check-then-act: two concurrent \`credit-low\` events would both read \"no recent email\", both send, then both insert the cooldown row.

Wrap the read+claim in a Serializable transaction so at most one event wins the cooldown slot. The loser either sees the inserted row or aborts on serialization failure, both of which short-circuit before sending. The cooldown row is also inserted up front rather than gated on send success, which is the right tradeoff: an SMTP failure shouldn't unlock a re-spam window.

redoh:
## Summary

- Two concurrent \`credit-low\` events could each read \"no recent email\" before either inserted the cooldown row, then both sent the email.
- Wraps the cooldown read and insert in a Serializable transaction so only one event wins the slot. The loser short-circuits either via the inserted row or a serialization-failure exception.
- The cooldown row is now claimed before \`sendEmail\` runs (previously inserted only on successful send). An SMTP failure should not unlock a 24h re-spam window.

## Test plan

- [ ] Trigger two near-simultaneous \`credit-low\` events for the same org and confirm only one email is sent and exactly one \`credit_low_email\` transaction row is written
- [ ] Confirm a serialization failure in the cooldown tx is logged and produces no email
- [ ] Confirm that after 24h the next \`credit-low\` event for the org sends again

Closes #378

### DIFF
--- a/apps/web/lib/events/observers/email.observer.ts
+++ b/apps/web/lib/events/observers/email.observer.ts
@@ -148,17 +148,46 @@ async function getAdminRecipients(
 const CREDIT_LOW_COOLDOWN_MS = 24 * 60 * 60 * 1000; // 24h
 
 async function onCreditLow(event: CreditLowEvent): Promise<void> {
-  // DB-backed cooldown: check last credit_low_email transaction
-  const recent = await prisma.creditTransaction.findFirst({
-    where: {
-      organizationId: event.orgId,
-      type: "credit_low_email",
-      createdAt: { gte: new Date(Date.now() - CREDIT_LOW_COOLDOWN_MS) },
-    },
-    select: { id: true },
-  });
+  // Atomically claim the cooldown slot before sending. Two concurrent
+  // `credit-low` events would otherwise both miss the cooldown read and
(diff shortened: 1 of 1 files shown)


redoh:
Re: "Serializable transaction relies on an unindexed filter column (`type`)" — acknowledging but not adding the composite `(organizationId, type, createdAt)` index in this PR. The `credit-low` cooldown path runs at most once per organization per 24h, so contention is essentially zero and predicate-lock coarseness has no measurable practical impact. Index addition involves a separate migration and is out of scope for the race-fix; happy to do it as a follow-up if data shows it's warranted.

redoh:
@octopus updates pushed (f1ba176) — serialization-failure log promoted to `console.error` and added an inline comment explaining that holding the 24h cooldown across SMTP failures is deliberate (preventing a re-spam loop). Ready for re-review.

PR #392, opened 2026-06-01
fix(review): verify "missing X" findings against full file to kill truncated-diff false positives

redoh:
## Problem

User feedback: the review reported "server route still missing", but the route actually existed, was wired correctly, and was tested. Root cause: a **truncated diff**.

In the two-pass validation, the second pass (`validateFindings`) cut the diff with a blind `slice(0, 12000)`:
- No truncation marker, so the model treated the 12k window as complete
- The file-tree ground truth was not passed to this pass
- "Missing X" claims were scored against that window alone

Result: when the route addition fell outside the window, the layer meant to **remove** false positives instead **confirmed** them by raising confidence.

## Fix

Verify existence ("missing X") claims against the **real file** instead of the diff — the only ground truth that survives truncation.

- "missing/absent/no/lacks X" findings are flagged as existence checks and carry the claimed symbol
- For existence claims, the **full file** is fetched and searched for the symbol, producing a definitive `>> SYMBOL "X" FOUND at file:L42` / `NOT FOUND` signal (instead of the previous 2000-char header)
- Added a truncation marker to the 12k validation window
- The repository file tree is now passed into the validation pass as existence ground truth
- Existence-specific scoring: symbol FOUND -> confidence 0-10 (false positive), fetch failed/unverifiable -> <=20, never raise confidence solely because something is absent from the diff window
- Per-finding file fetches are deduped via a per-call cache

## Scope note

Full-file existence fetch is only active for PR reviews (GitHub/GitLab/Bitbucket `fileContentFetcher`). The local review path (`review-core.ts`) calls validation without a fetcher; the truncation marker and file-tree guards still apply there, but full-file fetch does not. Wiring a local fetcher can be a separate task.

## Tests

- `tsc --noEmit`: 0 errors
- `reviewer.test.ts`: 58/58
- New `review-verification.test.ts`: 2/2 (existence-marking contract)

### DIFF
--- a/apps/web/lib/__tests__/review-verification.test.ts
+++ b/apps/web/lib/__tests__/review-verification.test.ts
@@ -9,7 +9,7 @@ function finding(overrides: Partial<InlineFinding>): InlineFinding {
     filePath: "apps/web/app/api/foo/route.ts",
     startLine: 1,
     endLine: 1,
-    category: "correctness",
+    category: "Bug",
     description: "",
     suggestion: "",
     confidence: 80,
@@ -21,8 +21,8 @@ describe("generateVerificationQueries — existence claims", () => {
   it("marks 'missing X' findings as existence checks with the symbol", () => {
     const findings = [
       finding({
-        title: "Server route still missing",
-        description: "The POST handler is missing from the route file.",
+        title: "Webhook route not registered",
+        description: "The route file is missing registerWebhook handler registration.",
       }),
     ];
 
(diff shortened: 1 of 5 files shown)


redoh:
@octopus updates pushed (`0d3d1e2`) — ready for re-review.

These findings landed in the review summary (no inline threads to react to / resolve), so here is the per-finding verdict:

### Addressed ✅

- **🔵 Test category uses non-standard value** (`review-verification.test.ts`) — Valid. Changed the fixture from `"correctness"` to the canonical `"Bug"` category.
- **💡 MISSING_PATTERN may extract a stop-word as the symbol** (`review-helpers.ts`) — Valid and important. A phrase like "missing **the** handler" made the lazy regex capture `the`, which always matches in a full-file search and would have wrongly suppressed a genuine missing-X finding. Added an `isUnusableSymbol()` guard (stop-word set + min length); such matches now fall through to the generic per-file context query. Also DRY'd the existing inline stop-word list into a shared `STOP_WORDS` set, and added a regression test.

### False positives ❌ (not changed)

- **🟡 DIFF_WINDOW constant duplicated / out of sync with slice** — There is no leftover hard-coded `diff.slice(0, 12000)`; the only slice uses `DIFF_WINDOW`, and the prompt template uses `${diffWindow}`. The "old" value the finding refers to is the **deleted** line visible in the diff (the `-` side), not code in the file. This is exactly the truncated/confused-diff misread this PR is meant to reduce.
- **🟡 buildExistenceEvidence context window can underflow** — No underflow is possible: `Math.max(0, ln - 4)` clamps the start and `Math.min(lines.length, ln + 5)` clamps the end. Overlapping snippets for hits within ~9 lines just repeat a few context lines with accurate line numbers; harmless to the validator.
- **💡 fileTree vs fileTreeStr naming inconsistency** — These are two independent local variables in two different files/scopes, not a shared symbol. There is no actual inconsistency to fix; renaming for cross-file cosmetic symmetry would be noise.

`tsc --noEmit`: 0 errors. Tests: 61/61 (incl. new stop-word regression test).

PR #511, opened 2026-06-10
fix(github): truncate PR comment bodies to avoid 422 on long reviews

cemoso:
## Problem

GitHub's issue / PR comment API rejects bodies > 65,536 chars with `422 Validation Failed (field=body, reason=too_long)`. The Octopus reviewer can produce review bodies above that cap when the model is verbose or the PR is large — the post then fails, and the LLM spend behind the failed call is already paid.

Observed live on a self-hosted deployment using `claude-fable-5`: a tight retry loop where every review costs ~\$1 and lands in `[reviewer] Review failed: Failed to update PR comment: 422`. Twelve in-flight PRs were unable to receive a review for over an hour.

## Fix

Add `truncateForGithubComment(body)` at the `apps/web/lib/github.ts` API boundary. Both `createPullRequestComment` and `updatePullRequestComment` route through it before sending.

- Cap: `MAX_GITHUB_COMMENT_BODY = 64_000` (~1.5k headroom for the marker).
- Prefer cutting at the nearest `\n\n` (or `\n## `) above 80% of the budget so truncation doesn't bisect a code fence or finding section.
- Appends `> ⚠️ Comment truncated — visit the dashboard for the full version.` so reviewers know there's more.
- Guard against ending on a lone high UTF-16 surrogate (the bot's findings are full of 🔴🟠🟡 severity emoji); `JSON.stringify` would otherwise emit an unpaired `\\ud83d` escape that GitHub may reject.

GitHub measures characters (Ruby code points); JS `.length` counts UTF-16 code units (always ≥ code points). So a 64,000-unit JS cap is strictly conservative against the 65,536-char API limit. Sources: [community#27190](https://github.com/orgs/community/discussions/27190), [renovate#14551](https://github.com/renovatebot/renovate/issues/14551).

## Scope

GitHub only — GitLab notes accept 1MB (review path is bounded by `maxTokens: 8192` so they're safe today), Bitbucket comments cap at 32KB but the failure-mode is currently latent (no retry loop). The same pattern can be ported to those providers in a follow-up.

## Test plan

- [x] `bun test apps/web/lib/__tests__/github-comment-truncate.test.ts` — 6 pass / 0 fail.
- [x] `bun run --cwd apps/web typecheck` clean.
- [x] Reviewed by an independent code-review pass (boundary, no-newline fallback, surrogate-pair safety, paragraph-boundary preference, JSON round-trip).
- [ ] Production: deploy fixes the running 422 loop on self-hosted instances.

## Diff size
2 files, +78 lines (1 new helper, 1 new test file).

### DIFF
--- a/apps/web/lib/__tests__/github-comment-truncate.test.ts
+++ b/apps/web/lib/__tests__/github-comment-truncate.test.ts
@@ -41,4 +41,17 @@ describe("truncateForGithubComment", () => {
     expect(out.length).toBeLessThanOrEqual(MAX_GITHUB_COMMENT_BODY);
     expect(out).toContain("Comment truncated");
   });
+
+  it("does not end on a lone surrogate (would break JSON encode)", () => {
+    // 🔴 is a surrogate pair in UTF-16. A body of contiguous 🔴 with no
+    // newlines lands in the hard-cut path; without the guard, slice may
+    // cut between the high and low surrogate and emit a malformed string.
+    // String.prototype.isWellFormed() (ES2024) reports unpaired surrogates.
+    const body = "🔴".repeat(MAX_GITHUB_COMMENT_BODY);
+    const out = truncateForGithubComment(body);
+    expect(out.length).toBeLessThanOrEqual(MAX_GITHUB_COMMENT_BODY);
+    expect(out.isWellFormed()).toBe(true);
+    // Sanity: re-encoding through JSON preserves all code points.
+    expect(() => JSON.parse(JSON.stringify({ body: out }))).not.toThrow();
+  });
 });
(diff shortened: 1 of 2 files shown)


cemoso:
Pushed `eb1e2d9` addressing both review points:

**🟡 `createPullRequestReview` body is not truncated** — fixed. The function now runs `body` through `truncateForGithubComment` before sending. This was the third call site I missed — the standard-pipeline review path goes through the reviews endpoint, not the issue-comments endpoint, so without this the bot would still 422 on finding-heavy reviews even though the issue-comment path was already safe.

**Code Quality 4/5 — `lastBoundary` may be -1** — clarified. Added an explicit `lastBoundary >= 0 &&` guard plus a comment explaining the sentinel. The numeric behavior is unchanged (`-1 > 0.8 * room` was already false), but intent is now visible from the expression.

Bitbucket/GitLab paths (the Consistency 4/5 note) intentionally left for a separate PR — those callers use different code paths that don't share the GitHub helper, and treating them in this PR would risk scope creep into platforms not relevant to the immediate outage.

6/6 existing truncation tests still pass.

@octopus please re-review.

johanfleming (approved):

PR #520, opened 2026-06-11
fix(self-host): pass user .env into the web container via env_file

cemoso:
## The bug

A fresh self-hoster following the quickstart writes their secrets (`BETTER_AUTH_SECRET`, OAuth client secrets, AI provider keys, integration credentials) into the root `.env`, then runs `docker compose up -d`. The web container starts with **none of those values**.

The current `docker-compose.yml`'s web service has only:

\`\`\`yaml
environment:
  - DATABASE_URL=postgresql://octopus:octopus@postgres:5432/octopus
  - QDRANT_URL=http://qdrant:6333
  - ENABLE_REVIEW_WORKERS=true
\`\`\`

No `env_file:` block. Compose uses the root `.env` file only for **variable substitution inside the compose file itself** — `${VAR}` references in compose YAML — not for delivering env vars to running containers. Containers only get keys that are explicitly listed under `environment:` or pointed at via `env_file:`.

## Impact

In practice, first-time self-hosters following the quickstart hit:

- `BETTER_AUTH_SECRET` unset → Better Auth throws at first session create → first sign-in 500s.
- GitHub App credentials unset → webhook routes 401, reviews never start.
- AI provider keys unset → first review fails with "no API key configured for org".

With `NODE_ENV=production`, the failures are silent/cryptic enough that a self-hoster typically gives up rather than diagnosing what's wrong.

## Fix

\`\`\`yaml
env_file:
  - path: .env
    required: false
\`\`\`

`required: false` keeps compose smoke-tests (no `.env` present) booting without changing today's failure mode for production users who **do** have a properly populated `.env`. The `environment:` block still overrides compose-internal `DATABASE_URL` / `QDRANT_URL` / `ENABLE_REVIEW_WORKERS` so the in-network service names take precedence over any same-name keys the user might have set.

## Test plan
- [ ] Smoke: `docker compose up -d` with NO .env file present — services start (today's behavior preserved).
- [ ] Quickstart end-to-end: populate .env with BETTER_AUTH_SECRET + DB seeds, `docker compose up -d`, sign in successfully on first try.
- [ ] Variable-substitution behavior in other services (e.g. `${OPENAI_API_KEY}` references in compose itself) still works — compose-side `.env` reading is independent of per-service `env_file:`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -12,10 +12,12 @@ services:
     # no AI provider keys, no integration credentials — they live in the
     # root .env which compose uses only for variable substitution in this
     # file by default, NOT for delivering env vars to running containers.
-    # required:false so smoke-tests with no .env file still boot.
+    # `required: true` (the default) means a missing .env fails compose
+    # explicitly rather than booting silently with no secrets — which is
+    # better than the prior failure mode where the container started but
+    # 500'd on the first auth call.
     env_file:
-      - path: .env
-        required: false
+      - .env
     environment:
       - DATABASE_URL=postgresql://octopus:octopus@postgres:5432/octopus
       - QDRANT_URL=http://qdrant:6333
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,17 @@ services:
       dockerfile: apps/web/Dockerfile
     ports:
       - "43300:3000"
+    # `env_file` is what passes the user's .env INTO the container; the
+    # `environment:` block below only sets the compose-internal DB / Qdrant
+    # URLs that override what the user might have set. Without this block,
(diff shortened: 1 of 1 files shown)


cemoso:
Pushed `9da1d59` addressing the 3/5 Security finding:

You were right — `required: false` re-introduced the exact failure mode this PR is supposed to fix. Dropped the knob (default is now `required: true`). A missing .env now fails compose explicitly at startup with a clear "env file ./.env not found" message instead of booting silently and 500'ing at the first auth call.

The "smoke tests want to boot without secrets" use case can be served by `touch .env` — explicit empty file, zero ambiguity about whether the user meant to skip secrets vs forgot them.

@octopus please re-review.

johanfleming (approved):

PR #521, opened 2026-06-11
fix(embeddings): assert returned vector dim matches Qdrant collection

cemoso:
## The bug

`getEmbedModel(orgId, repoId)` returns a per-repo / per-org / platform-default model string. The Qdrant `code_chunks` collection is configured for `VECTOR_SIZE = 3072` (text-embedding-3-large's native dim). If an org sets `defaultEmbedModelId` (or a repo sets `embedModelId`) to anything else — `text-embedding-3-small` (1536), `text-embedding-ada-002` (1536), some Azure deployments — the embeddings layer happily produces vectors at the OVERRIDDEN model's dim, then Qdrant 400s at upsert time with a cryptic "vector dim mismatch."

The 400 fires minutes after the user has triggered a re-index, after the LLM bill for thousands of chunks has already been paid. The error message is also opaque to anyone who didn't configure the override personally.

## Fix

Two small changes:

### 1. Request the right dim where the API supports it

\`\`\`ts
const supportsDimensionsParam = /^text-embedding-3-/.test(embedModel);

const res = await client.embeddings.create({
  model: embedModel,
  input: batch,
  ...(supportsDimensionsParam ? { dimensions: EMBED_VECTOR_SIZE } : {}),
});
\`\`\`

OpenAI's text-embedding-3-* family accepts a `dimensions:` parameter that truncates the response to the requested size. Passing it means an `embedModelId` override to text-embedding-3-small produces 3072-dim vectors compatible with the existing collection.

### 2. Validate the response regardless

\`\`\`ts
if (item.embedding.length !== EMBED_VECTOR_SIZE) {
  throw new Error(
    `Embedding model "\${embedModel}" returned \${item.embedding.length}-dim vectors but the Qdrant collection is configured for \${EMBED_VECTOR_SIZE}. Either pick a model whose native or requested dim matches the collection (text-embedding-3-large @ 3072 by default), or re-create the Qdrant collection at the model's dim.`,
  );
}
\`\`\`

text-embedding-ada-002 and many custom Azure deployments **ignore** the `dimensions:` parameter — they always return their native dim. Without the assert, you'd still hit the cryptic Qdrant 400. With it, the failure surfaces immediately with an actionable message naming the model and both dims, before any LLM money is spent on the rest of the batch.

`VECTOR_SIZE` was a private const in `lib/qdrant.ts`. Exposed it as `EMBED_VECTOR_SIZE` (preserving the local symbol for the dense set of existing call sites in qdrant.ts itself) so the embeddings layer can reference the canonical value instead of duplicating the magic number.

## Test plan
- [ ] Production: temporarily set an org's `defaultEmbedModelId` to `text-embedding-3-small`, trigger re-index — vectors come back at 3072-dim and Qdrant accepts them.
- [ ] Set to `text-embedding-ada-002` (no `dimensions:` support, native 1536) — get the new actionable error message on the first batch instead of waiting for Qdrant.

### DIFF
--- a/apps/web/lib/__tests__/embeddings-dim-check.test.ts
+++ b/apps/web/lib/__tests__/embeddings-dim-check.test.ts
@@ -1,0 +1,104 @@
+import { describe, it, expect, mock, beforeEach } from "bun:test";
+import { QDRANT_DENSE_VECTOR_SIZE } from "@/lib/qdrant";
+
+// `server-only` is a Next.js marker module that throws if imported from a
+// non-server context. The test runner is "non-server" enough to trip it.
+mock.module("server-only", () => ({}));
+
+// Mock the AI/usage/cost helpers — none of them are exercised by the
+// assertion path we're testing.
+mock.module("@/lib/cost", () => ({
+  isOrgOverSpendLimit: () => Promise.resolve(false),
+}));
+mock.module("@/lib/ai-usage", () => ({
+  logAiUsage: () => Promise.resolve(),
+}));
+
+// Returned embedding dim is the lever — tests reset it per case.
+let mockReturnedDim = QDRANT_DENSE_VECTOR_SIZE;
+
+mock.module("@/lib/ai-client", () => ({
+  // No tracking → embeddings uses "text-embedding-3-large" default; with
+  // tracking it asks ai-client. Cover both paths by always returning the
+  // override model the test cares about.
+  getEmbedModel: () => Promise.resolve("text-embedding-3-small"),
+}));
+
+mock.module("openai", () => {
+  // Minimal stub matching the shape embeddings.ts uses:
+  // `new OpenAI(...)` + `.embeddings.create(...)` returning `{ data, usage }`.
+  // embeddings.ts also references `OpenAI.APIError` (static) in its catch
+  // branch — must be present as a class, otherwise `err instanceof X` throws
+  // "Right hand side of instanceof is not an object" before reaching the
+  // dim-check assertion we want to test.
+  class StubAPIError extends Error {
+    status?: number;
+  }
+  class StubOpenAI {
+    static APIError = StubAPIError;
+    embeddings = {
+      create: ({ input }: { input: string[] }) =>
+        Promise.resolve({
+          data: input.map(() => ({ embedding: new Array(mockReturnedDim).fill(0.1) })),
+          usage: { prompt_tokens: 1 },
+        }),
+    };
+  }
+  return { default: StubOpenAI, APIError: StubAPIError };
+});
+
+// Import AFTER mocks are wired so module init picks them up.
+const { createEmbeddings } = await import("@/lib/embeddings");
+
+describe("createEmbeddings — Qdrant dim validation", () => {
+  beforeEach(() => {
+    mockReturnedDim = QDRANT_DENSE_VECTOR_SIZE; // default: matches
+  });
+
+  // tracking arg forces `embedModel = await getEmbedModel(orgId, repoId)`
+  // (the mock returns "text-embedding-3-small"). Without tracking,
+  // createEmbeddings hardcodes the default model — not the path this PR
+  // is hardening.
+  const TRACKING = { organizationId: "org-test", operation: "index-test" };
+
+  it("passes through when returned vectors match collection dim", async () => {
+    mockReturnedDim = QDRANT_DENSE_VECTOR_SIZE;
+    const out = await createEmbeddings(["hello", "world"], TRACKING);
+    expect(out.length).toBe(2);
+    expect(out[0].length).toBe(QDRANT_DENSE_VECTOR_SIZE);
+    expect(out[1].length).toBe(QDRANT_DENSE_VECTOR_SIZE);
+  });
+
+  it("throws an actionable error when returned dim is smaller than collection (e.g. text-embedding-3-small native)", async () => {
+    mockReturnedDim = 1536; // text-embedding-3-small / ada-002 native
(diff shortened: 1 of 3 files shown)


cemoso:
Pushed `b8317a6` addressing all three findings:

**🟠 EMBED_VECTOR_SIZE only reflects code_chunks** — fixed. Renamed to `QDRANT_DENSE_VECTOR_SIZE` and expanded the doc comment to explicitly enumerate all 7 collections this module creates (code_chunks, knowledge, review, chat, diagram, feedback, docs — verified: all pass \`size: VECTOR_SIZE\` to createCollection). Added a paragraph documenting what to do if a future collection ever uses a different dim (decouple the constant + pass expected dim as arg to embeddings layer).

**🟡 Partial validVectors after mid-batch failure** — fixed. Restructured embedBatch to validate ALL items in a batch first, then push them all. A mid-batch dim mismatch now leaves validVectors completely untouched for the failed batch.

**💡 Circular-dependency risk** — investigated. \`qdrant.ts\` does NOT import from \`embeddings.ts\` (grepped). The export is a one-way dependency: qdrant.ts owns the constant, embeddings.ts consumes it. No cycle exists today. Not addressing — leaving the constant where it lives.

@octopus please re-review.

redoh:
Thanks for this — the dim-mismatch assert (fix 2) is a real improvement and worth merging. But fix 1 (the `dimensions:` parameter) doesn't work the way the PR describes, and it should be removed before merge.

## The `dimensions:` parameter can only truncate, never expand

OpenAI's `dimensions` parameter performs dimensionality *reduction* (MRL truncation + renormalization). It cannot request more dimensions than the model's native size:

- `text-embedding-3-small` is natively 1536-dim. Requesting `dimensions: 3072` does not produce 3072-dim vectors — the API rejects the request with a 400 `invalid_request_error` ("dimensions must be less than or equal to 1536").
- So the headline claim — "an `embedModelId` override to text-embedding-3-small produces 3072-dim vectors compatible with the existing collection" — is incorrect. The first unchecked test-plan item would fail if run.

## Worse: fix 1 preempts fix 2's actionable error

For any `text-embedding-3-*` override with a native dim below 3072, the OpenAI 400 now fires *before* your new assert ever runs. The user gets OpenAI's generic parameter error instead of the actionable message you wrote naming the model, both dims, and the two remediation options. Without the `dimensions:` parameter, the same misconfiguration flows into the assert and produces the better error.

## The parameter has no useful case in this codebase

The Qdrant collections are all at 3072, which is `text-embedding-3-large`'s native dim — the largest of any supported model. Truncation *down* to the collection size can therefore never apply:

- `text-embedding-3-large` → `dimensions: 3072` is a no-op (native size).
- `text-embedding-3-small` / anything smaller → API error, as above.
- `ada-002` / Azure deployments → already excluded by the `supportsDimensionsParam` guard.

## Requested change

Drop the `dimensions:` parameter and the `supportsDimensionsParam` guard entirely; keep the assert (including the whole-batch-validate-before-push structure — that part is right). The assert alone delivers the actual value of this PR: fail fast with an actionable message before the LLM bill is paid.

Would also be good to add a small unit test for the assert path (mock a response with wrong-dim vectors, expect the actionable error), since the PR currently ships without tests.

redoh (changes requested):
Requesting changes per my comment above (https://github.com/octopusreview/octopus/pull/521#issuecomment-4680956315) — TL;DR:

1. **Drop the `dimensions:` parameter and the `supportsDimensionsParam` guard.** OpenAI's `dimensions` can only truncate, never expand: `text-embedding-3-small` (native 1536) rejects `dimensions: 3072` with a 400, so the "small model now produces collection-compatible vectors" claim doesn't hold. Worse, that 400 fires before the new assert, replacing your actionable error with OpenAI's generic one. Since the collection dim (3072) equals the largest model's native dim, the parameter has no useful case here.

2. **Keep the assert** (fix 2) exactly as is, including validate-whole-batch-before-push — that's the real value of this PR.

3. **Add a unit test for the assert path** (mock a wrong-dim response, expect the actionable error); the PR currently ships without tests.

cemoso:
@redoh thank you — you're right, I had the `dimensions:` parameter backwards. Pushed `deb949f` with all three of your asks:

1. **Dropped `dimensions:` and the `supportsDimensionsParam` guard.** Verified your reasoning: collection is 3072 = text-embedding-3-large's native, so truncation-down never applies; for text-embedding-3-small the 400 was preempting the actionable error. Now the request is just `{ model, input }` again and any size mismatch flows into the assert as you described.

2. **Kept the assert** verbatim, including validate-whole-batch-before-push.

3. **Added `apps/web/lib/__tests__/embeddings-dim-check.test.ts`** — 4 cases:
   - pass-through on matching dim
   - throw on under-dim (text-embedding-3-small / ada-002)
   - error message contains model name + both dims + remediation guidance
   - whole-batch atomicity (4-item batch fails once, no partial state)

4/4 pass.

@octopus please re-review.

redoh (approved):
All three asks from the change request are addressed in `deb949f`:

1. `dimensions:` parameter and the `supportsDimensionsParam` guard are gone — the request is back to `{ model, input }`, and any dim mismatch now flows into the assert with the actionable message.
2. The assert is kept verbatim, including validate-whole-batch-before-push.
3. 4 new unit tests cover the assert path, including the error-message content and the atomic mid-batch failure contract. The `OpenAI.APIError` stub detail is a nice touch.

Verified locally: the new dim-check tests pass 4/4 and the full suite is green (312/312) on this branch. Approving.

johanfleming (approved):

PR #525, opened 2026-06-11
feat(reviewer): three anti-hallucination fields on every finding

cemoso:
## What

Three required fields added to the finding JSON schema in `apps/web/prompts/SYSTEM_PROMPT.md`:

| Field | Purpose |
|---|---|
| `whyTestsDoNotAlreadyCoverThis` | Names the specific tests inspected and why they fail to exercise this case. If no concrete test gap can be named, the finding is speculative — downgrade confidence or skip. |
| `suggestedRegressionTest` | A concrete test that would fail today and pass once the suggestion is applied. |
| `minimumFixScope` | The smallest change that resolves the issue. Discourages cross-cutting refactors disguised as bug fixes. |

Plus rule **12a** in the review checklist that anchors `whyTestsDoNotAlreadyCoverThis` as the structural anti-hallucination guard: if you can't name a concrete test gap, the finding is speculative.

## Why now

The current schema asks the model to be confident but doesn't make it do the work that grounds confidence. Two failure modes today:

1. **Speculative findings on covered code** — the model flags a bug that an existing test already proves doesn't exist. The reviewer reads the inline comment, looks at the test file, bins the finding as noise.
2. **Cross-cutting refactor disguised as a bug fix** — a 1-line bug gets a 12-file "suggested fix" that requires substantial review effort beyond verifying the actual bug.

Both failure modes are structural, not prompt-engineering. The new fields make the model DO the verification (find tests covering the line, write the regression test) and record the answer in JSON. Forcing the answer into the field discourages the speculative finding entirely — there's no way to fill `whyTestsDoNotAlreadyCoverThis` honestly for a finding the existing tests already disprove.

## Compatibility

- `parseFindingsFromJson` in `apps/web/lib/review-dedup.ts` validates only the required fields it consumes (`severity`, `title`, `filePath`, `startLine`, `description`). Extra fields are accepted and silently ignored. No parser change needed for this PR to ship.
- Older reviews (without these fields) continue to parse and render.
- Inline comments still render via the existing `suggestion` field. The new fields appear in the JSON code block in the review body for anyone wanting to audit the model's grounding work.

## Risk

Pure prompt change. No code modified, no tests modified, no schemas changed. Worst case: the model occasionally fills the new fields with vague answers — the existing review experience is unchanged because nothing downstream depends on them.

## Test plan
- [ ] Trigger a review on this PR itself — confirm Octopus's own output includes the three new fields and the answers read sensibly.
- [x] Verified parser tolerance: `parseFindingsFromJson` only checks the five required fields it consumes; extras pass through.

### DIFF
--- a/apps/web/lib/review-dedup.ts
+++ b/apps/web/lib/review-dedup.ts
@@ -15,6 +15,15 @@ export type InlineFinding = {
   description: string;
   suggestion: string;
   confidence: number;
+  // Anti-hallucination fields — see apps/web/prompts/SYSTEM_PROMPT.md.
+  // Optional in the type because (a) older review bodies don't contain
+  // them and the parser is intentionally permissive about that, and
+  // (b) the prompt instructs the model to always include them, so when
+  // present they're reflected here for downstream surfaces (UI, audit
+  // export) without re-parsing the raw JSON block.
+  whyTestsDoNotAlreadyCoverThis?: string;
+  suggestedRegressionTest?: string;
+  minimumFixScope?: string;
 };
 
 export type PriorFinding = {
@@ -93,6 +102,18 @@ export function parseFindingsFromJson(reviewBody: string): InlineFinding[] | nul
             : item.confidence === "HIGH"
               ? 90
               : 70,
+        // Anti-hallucination fields — typed only when the model emitted a
+        // string. Absent fields stay undefined so the existing UI surfaces
+        // continue to render exactly as before for older reviews.
+        ...(typeof item.whyTestsDoNotAlreadyCoverThis === "string"
+          ? { whyTestsDoNotAlreadyCoverThis: item.whyTestsDoNotAlreadyCoverThis }
+          : {}),
+        ...(typeof item.suggestedRegressionTest === "string"
+          ? { suggestedRegressionTest: item.suggestedRegressionTest }
+          : {}),
+        ...(typeof item.minimumFixScope === "string"
+          ? { minimumFixScope: item.minimumFixScope }
+          : {}),
       });
     }
 
--- a/apps/web/lib/review-helpers.ts
+++ b/apps/web/lib/review-helpers.ts
@@ -423,6 +423,23 @@ export function buildInlineComments(
       body += `\n\n${suggestionBlock}`;
     }
 
+    // Surface the model's grounding work to the reader. These come from
+    // the anti-hallucination fields in the JSON schema (see SYSTEM_PROMPT.md):
+    //   - minimumFixScope keeps the suggested fix small and reviewable
+    //   - suggestedRegressionTest gives the dev a concrete test to add
+    // Without this block the fields only survived in the DB row — the
+    // tokens spent on them never reached the user.
+    if (f.minimumFixScope || f.suggestedRegressionTest) {
+      const groundingParts: string[] = [];
+      if (f.minimumFixScope) {
+        groundingParts.push(`**Minimum fix scope:** ${f.minimumFixScope}`);
+      }
+      if (f.suggestedRegressionTest) {
+        groundingParts.push(`**Suggested regression test:**\n\`\`\`\n${f.suggestedRegressionTest}\n\`\`\``);
+      }
+      body += `\n\n<details><summary>📌 Grounding</summary>\n\n${groundingParts.join("\n\n")}\n\n</details>`;
+    }
+
     // AI Fix Prompt — collapsible section with copy-pasteable prompt
     const severityLabel = f.severity === "🔴" ? "Critical" : f.severity === "🟠" ? "High" : f.severity === "🟡" ? "Medium" : f.severity === "🔵" ? "Low" : "Nit";
     const categoryNote = f.category ? ` (${f.category})` : "";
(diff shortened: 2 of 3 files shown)


cemoso:
@octopus please re-review — addressed the two 🟡s, left the 💡 verbose-name nit (the verbosity is the point: unambiguous signal that this is a runnable test, not prose).

cemoso:
@octopus please re-review — addressed the parser/prompt consistency gap. Added the three fields to the `InlineFinding` TypeScript type as optional + threaded them through `parseFindingsFromJson` via conditional spread. Permissive on absence by design (older review bodies don't carry them and shouldn't drop on re-render). 42/42 existing review-dedup tests still pass.

redoh:
Took a closer look at this with an eye on how the new fields interact with the rest of the pipeline. The direction is good and the risk is low, but a few caveats worth addressing (or at least acknowledging) before/after merge:

### 1. The new fields are currently not visible anywhere on GitHub

The PR description says the fields "appear in the JSON code block in the review body for anyone wanting to audit the model's grounding work", but that is not the case:

- `stripDetailedFindings` (`apps/web/lib/review-helpers.ts:265`, the #511/#512 fix) removes the JSON findings block from the main comment entirely.
- `buildInlineComments` (`apps/web/lib/review-helpers.ts:417`) only renders `description` + `suggestion` (+ the AI Fix Prompt), and ignores the three new fields.

So the fields only survive in the `reviewBody` column in the DB. The grounding effect still works (the model has to do the verification either way), but the tokens spent on `suggestedRegressionTest` never reach the user. Suggested follow-up: surface `suggestedRegressionTest` and `minimumFixScope` in the inline comment or inside the AI Fix Prompt block.

### 2. Output token cost and truncation risk

Every finding now carries three extra fields, one of which is a code snippet. On findings-heavy reviews this meaningfully grows the output. The sharp edge: if the JSON output gets cut off at `max_tokens`, `parseFindingsFromJson` returns `null` on the parse error, the markdown fallback finds nothing, and the entire review loses all findings. The floor raised in 9e61ac3 helps, but this change pushes outputs closer to that limit. Worth monitoring after deploy.

### 3. Can the model actually see the tests it is asked to cite?

`whyTestsDoNotAlreadyCoverThis` instructs the model to name "the specific tests you inspected". If the relevant test files do not make it into the review context (diff or retrieval), the model cannot answer honestly and may fabricate test file paths, which would make the anti-hallucination field itself a hallucination source. Suggest spot-checking the cited test paths in the first few real reviews after deploy.

### 4. The main test-plan item is still unchecked

The Octopus reviews on this PR ran with the currently deployed prompt, not this branch's prompt, so "confirm Octopus's own output includes the three new fields" can only happen after merge + deploy. Fine to merge as is, but the real validation is post-deploy. Also: if octopus-engine carries its own copy of the prompt, it will drift from this one.

### 5. Possible suppression of true positives

"If you cannot name a concrete test gap, downgrade or skip" may make the model overly conservative on well-tested repos and drop real findings. Not a merge blocker, but worth watching the findings-per-review trend after this lands.

None of these block the merge: it is a pure prompt + optional-type change and trivially revertible. Points 1 and 3 are the ones I would act on.

cemoso:
@redoh thank you for the close read — really appreciated. Pushed `25c2797` acting on point 1; acknowledging the others below.

### 1. Fields not visible anywhere — ADDRESSED
You were right that the tokens never reached the user. Added a collapsed `📌 Grounding` section to the inline comment between the suggestion block and the AI Fix Prompt:
- `minimumFixScope` as a one-liner
- `suggestedRegressionTest` as a fenced code block

Also folded `minimumFixScope` into the AI Fix Prompt as a `Scope:` line — an assistant running that prompt should respect the bounded change scope instead of expanding into a cross-cutting refactor.

`whyTestsDoNotAlreadyCoverThis` is **intentionally not surfaced** for the reason in your point 3 below — until we can guarantee the cited tests are in the retrieval context, exposing the citation invites the reader to verify a path the model may have fabricated. It stays internal (in the DB row for audits, not in the public PR comment).

### 2. Output token cost and truncation risk — ACKNOWLEDGED, no action
Agreed this pushes outputs closer to the cap. The `findings.length > 0 ? findings : null` return shape in `parseFindingsFromJson` is the cliff. The `stripDetailedFindings` cap on comment body size (#511) catches the visible-comment side, but the *parse* side has no graceful degradation today. Worth a follow-up to either (a) bias the model toward shorter values in the new fields when finding count is high, or (b) parse partial JSON on truncation. Not blocking this PR; flagged for tracking.

### 3. Cited test paths may be hallucinated — ACKNOWLEDGED, partial mitigation
Right concern. Two mitigations:
- I deliberately don't surface `whyTestsDoNotAlreadyCoverThis` in the inline comment (per point 1 above), so a hallucinated test path doesn't reach the developer to be checked.
- Spot-check protocol on first reviews after deploy: pull a handful of `whyTestsDoNotAlreadyCoverThis` values from the DB and verify the cited paths exist. If hallucination rate is meaningful, follow up with explicit retrieval of `__tests__/<targetFile>.test.ts` into the review context.

### 4. Test-plan validation is post-deploy — ACKNOWLEDGED
True — the bot reviewing this branch ran with the deployed prompt, not this branch's. Real validation is post-deploy. Also flagging the octopus-engine-carries-its-own-copy drift risk in the project tracker.

### 5. Possible suppression of true positives — ACKNOWLEDGED, watching
"If you cannot name a concrete test gap, downgrade or skip" could over-suppress on heavily-tested code. Worth watching the findings-per-review trend after deploy. If it dips significantly, soften the rule from "downgrade or skip" to "downgrade only" so the finding still surfaces but at lower confidence.

@octopus please re-review once point 1's change is verified.

johanfleming (approved):

PR #526, opened 2026-06-14
feat(lib): robust JSON extraction with three-tier recovery

cemoso:
## What

Adds `apps/web/lib/extract-json.ts` — best-effort JSON extraction from LLM output, with three-tier recovery:

1. **Strict** — `JSON.parse` on the whole input.
2. **Code-fence** — strip a single triple-backtick block and parse the inner content.
3. **Balanced scan** — walk the input looking for balanced `[...]` or `{...}` blocks; string-literal aware so escaped quotes and braces inside strings don't desync the depth counter.

Array roots are scanned **before** object roots in tier 3. An unfenced findings array `[{...}, {...}]` would otherwise match the first inner `{` standalone and drop the rest of the array.

## Why

Two call sites in the codebase previously rolled their own (incomplete) version of tiers 1 + 2:

| Caller | Failure mode without tier 3 |
|---|---|
| `lib/review-dedup.ts::parseFindingsFromJson` | Inline implementation handled only `JSON.parse(stripped-fence)`. If the model emitted any prose between the `OCTOPUS_FINDINGS_START`/`_END` markers (a common failure mode under output-token pressure), the parse returned null and **every finding dropped silently**. |
| `lib/summarizer.ts::summarizeRepository` | Same shape, same gap. An LLM that wrapped its `{purpose, summary}` in any prose ("Here's the summary:") fell through to the `text.slice(0, 500)` fallback even though the JSON itself was perfectly valid. |

Both now use the shared helper. The balanced-scan path recovers prose-wrapped JSON in both cases.

## Test plan

- [x] 18 new unit tests in `apps/web/lib/__tests__/extract-json.test.ts`: tier-1 strict (object/array/non-JSON/whitespace), tier-2 fence (with/without language hint, unparseable inner), tier-3 balanced scan (unfenced object in prose, unfenced array in prose, braces in string literal, escaped quotes, skip-malformed-and-try-next, no-candidate-parses), `extractJsonObject` (object/array/primitive/null/embedded).
- [x] All 42 existing tests in `apps/web/lib/__tests__/review-dedup.test.ts` pass against the refactored `parseFindingsFromJson` — the new path is strictly a superset of the old one.

## Risk

Three-tier recovery never produces a parse where the old strict path also produced one. It only *adds* recovery paths. The worst plausible regression is the balanced-scan path producing a different (but still valid) JSON candidate from prose that the strict path would have rejected — and that's the entire point. Reverting the adoption (e.g. by going back to `JSON.parse(block)`) restores prior behavior exactly.

### DIFF
--- a/apps/web/lib/__tests__/extract-json.test.ts
+++ b/apps/web/lib/__tests__/extract-json.test.ts
@@ -72,6 +72,15 @@ describe("extractJson — tier 3 (balanced scan)", () => {
   it("returns null when no candidate parses", () => {
     expect(extractJson("definitely not json {nope nope}")).toBeNull();
   });
+
+  it("recovers a valid candidate after an unterminated string desyncs an earlier one", () => {
+    // A lone `"` inside the first candidate flips inString=true and is never
+    // toggled back, so the matching close-brace is treated as in-string and
+    // the candidate never balances. Scanner must advance to a later open
+    // delimiter rather than bailing entirely.
+    const input = 'first {"a": "unterminated string here ok ok ok    later {"b":1}';
+    expect(extractJson(input)).toEqual({ b: 1 });
+  });
 });
 
 describe("extractJsonObject", () => {
--- a/apps/web/lib/__tests__/extract-json.test.ts
+++ b/apps/web/lib/__tests__/extract-json.test.ts
@@ -1,0 +1,99 @@
+import { describe, it, expect } from "bun:test";
+import { extractJson, extractJsonObject } from "@/lib/extract-json";
+
+describe("extractJson — tier 1 (strict)", () => {
+  it("parses a clean JSON object", () => {
+    expect(extractJson('{"a":1,"b":"x"}')).toEqual({ a: 1, b: "x" });
+  });
+
+  it("parses a clean JSON array", () => {
+    expect(extractJson("[1,2,3]")).toEqual([1, 2, 3]);
+  });
+
+  it("returns null for non-JSON input", () => {
+    expect(extractJson("this is not json at all")).toBeNull();
+  });
+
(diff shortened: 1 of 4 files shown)


cemoso:
Pushed addressing both 🟡 findings:

🟡 **Unterminated string corrupts depth counting** — fixed. When the for-loop exits without depth reaching 0, the scanner now advances `firstOpen` by ONE rather than returning null. A later well-formed candidate elsewhere in the text still gets its chance. Added a regression test exercising exactly that case (first candidate has a lone `"`, second is well-formed — scanner returns the second).

🟡 **`extractJsonObject` null-check uses truthy short-circuit** — tightened. `parsed && ...` was correct (every non-null object is truthy in JS) but slightly opaque. Replaced with `parsed !== null && ...` — same semantics, more direct about excluding null (the only falsy "object" value).

19/19 tests pass.

@octopus please re-review.

redoh (approved):
Thanks for this, @cemoso — clean, well-motivated, and the test coverage is excellent. 🙏

I rebased the branch onto latest `master` to clear the merge conflict. The conflict was with #525, which had just landed three anti-hallucination fields (`whyTestsDoNotAlreadyCoverThis`, `suggestedRegressionTest`, `minimumFixScope`) on every finding in the exact `findings.push({...})` block this PR rewrites. I resolved it to keep **both**: the `extractJson` delegation from this PR *and* all three fields from #525 — they'd have been silently dropped in a naive "take ours" resolution.

Nothing else changed — `extract-json.ts`, its test suite, and the `summarizer.ts` migration are exactly as you wrote them. The PR is now mergeable.

Thanks again, especially for the adversarial cases like the unterminated-string desync — that's exactly the kind of coverage that keeps LLM-output parsing from regressing.

PR #534, opened 2026-06-19
docs(github-action): document fork pull request support

redoh:
## What changed

Adds a **Fork Pull Requests** section to the GitHub Action docs page explaining that `GITHUB_TOKEN` is read-only for fork PRs on the `pull_request` event, with:
- a `pull_request_target` recipe to enable reviews on fork PRs (safe here since the action only reads the diff and never checks out/runs fork code),
- a pointer to the GitHub App for repos that prefer not to grant a writable CI token,
- an FAQ entry linking to the new section.

`Section` now accepts an optional `id` so `#fork-pull-requests` anchors resolve.

## Why

Pairs with [octopusreview/action#1](https://github.com/octopusreview/action/pull/1), which surfaces the read-only-token failure in the action logs and links here. Context: a maintainer hit this silently on [kubernetes/minikube#23143](https://github.com/kubernetes/minikube/pull/23143) and closed the PR without knowing the cause.

## Validation

- `bun run lint` passes (2 pre-existing warnings in unrelated files).

### DIFF
--- a/apps/web/app/(landing)/docs/github-action/page.tsx
+++ b/apps/web/app/(landing)/docs/github-action/page.tsx
@@ -73,6 +73,23 @@ const pathsYaml = `on:
       - "lib/**"
 `;
 
+const forkYaml = `# .github/workflows/octopus.yml
+name: Octopus Review
+on:
+  pull_request_target:
+    types: [opened, synchronize]
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  review:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


redoh:
Thanks, good catch 🙏 Addressed in 96be898.

The `pull_request_target` checkout caveat was buried in prose. Added a prominent amber security callout directly below the workflow recipe, spelling out that checking out and running the fork's head commit in a `pull_request_target` context executes untrusted code with a writable token and secrets (a known RCE vector), and clarifying that the recipe above is safe because the action only reads the diff via the API.

@octopus updates are ready for re-review.

PR #538, opened 2026-06-23
feat(blog): add RSS feed at /blog/feed.xml

### LINKED ISSUE #496: Add an RSS feed for the blog
The blog has a sitemap but no RSS feed, so readers cannot subscribe to new posts.

### Acceptance criteria
- Add an RSS 2.0 (or Atom) feed route for published blog posts.
- Link it from the blog page `<head>`.

### Relevant files
- New: `apps/web/app/blog/feed.xml/route.ts` (or `/api/blog/feed`)
- Reference: `apps/web/app/sitemap.ts`

pushkarbanjare:
## Summary

fixes #496 

## Changes

- Adds an RSS 2.0 feed for the blog (#496).
- New /blog/feed.xml route that lists the most recent published posts.
- Discovery <link rel="alternate" type="application/rss+xml"> added to the blog page <head> via metadata.alternates.types.
- Mirrors sitemap.ts: same status: "published", deletedAt: null query.
- Unit tests for the feed builder included.

## Test plan

- [x] Linting passes (`bun run lint`)
- [x] Type check passes (`bun run typecheck`)
- [x] Tested locally

### DIFF
--- a/apps/web/app/(landing)/blog/feed.xml/route.ts
+++ b/apps/web/app/(landing)/blog/feed.xml/route.ts
@@ -1,0 +1,40 @@
+import { prisma } from "@octopus/db";
+import { buildRssFeed } from "@/lib/blog-rss";
+
+export const dynamic = "force-dynamic";
+
+const MAX_ITEMS = 20;
+
+function fetchPublishedPosts() {
+  return prisma.blogPost.findMany({
+    where: { status: "published", deletedAt: null },
+    orderBy: { publishedAt: "desc" },
+    take: MAX_ITEMS,
+    select: {
+      title: true,
+      slug: true,
+      excerpt: true,
+      authorName: true,
+      publishedAt: true,
+      updatedAt: true,
+    },
+  });
+}
+
+export async function GET() {
+  let posts: Awaited<ReturnType<typeof fetchPublishedPosts>> = [];
+  try {
+    posts = await fetchPublishedPosts();
+  } catch {
+    posts = [];
+  }
+
+  const xml = buildRssFeed(posts);
+
+  return new Response(xml, {
+    headers: {
+      "Content-Type": "application/rss+xml; charset=utf-8",
+      "Cache-Control": "public, s-maxage=3600, stale-while-revalidate=86400",
+    },
+  });
+}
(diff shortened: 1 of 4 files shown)


redoh:
@pushkarbanjare Thanks for the PR! Clean, well-tested addition. A manual pass surfaced a few small items that complement the automated review. None are blocking.

**1. `slug` is interpolated into `<link>`/`<guid>` without XML escaping** — `lib/blog-rss.ts`

Every other field (`title`, `excerpt`, `authorName`) goes through `escapeXml`, but `link` does not:

```ts
const link = `${SITE_URL}/blog/${post.slug}`;
```

Slugs are `@unique` and normally URL-safe, so the risk is low, but it's the only unescaped interpolation in the document. A slug containing `&` or `<` would produce invalid XML. Suggest `escapeXml(post.slug)` for consistency.

**2. `authorName` nullable branch is effectively dead code** — `lib/blog-rss.ts`

The type declares `authorName: string | null` and the code conditionally omits `<dc:creator>`, but the Prisma schema has `authorName String` (non-nullable). The `else` branch (and the `omits dc:creator when authorName is null` test) can never trigger in practice. Harmless, just over-defensive.

**3. (optional) Existing `escapeXmlAttr` helper** — `lib/repo-config-shared.ts`

There's already an XML-escaping helper in the repo. Different concern (repo config vs blog feed) and the new `escapeXml` is tested, so consolidating isn't required, just flagging the overlap.

---

Please also address the findings from the **Octopus Review** above before merging, in particular:

- 🟡 **Silent DB error swallows failures** (`route.ts`) — the `catch` resets `posts` to `[]` and still returns a 200 empty feed with no log. At minimum add a `console.error`; ideally return 503 + `Retry-After` so aggregators retry instead of caching an empty feed.
- 🔵 **`force-dynamic` vs `revalidate`** — `export const revalidate = 3600` matches the `s-maxage=3600` intent and lets Next.js cache instead of hitting the DB on every poll.
- 🔵 **Non-deterministic `lastBuildDate`** — the `new Date()` fallback changes on every request for an empty feed; a fixed/build-time fallback is more correct.

pushkarbanjare:
I have made the changes according to the suggestions, and do you think shall we move `escapeXml` into `lib/utils.ts` and import it to `lib/blog-rss.ts` and `lib/repo-config-shared.ts`?

redoh (approved):

redoh:
> I have made the changes according to the suggestions, and do you think shall we move `escapeXml` into `lib/utils.ts` and import it to `lib/blog-rss.ts` and `lib/repo-config-shared.ts`?

Thanks! No need to move it for now, two call sites don't justify a shared helper yet. We can extract it if a third one appears. Approved! 🚀

PR #541, opened 2026-06-26
fix(slack): bind OAuth state to initiating session (account-linking CSRF)

redoh:
## Summary

Fixes an OAuth account-linking CSRF in the Slack integration (reported privately under the bug bounty policy).

The Slack OAuth `state` was only base64-encoded JSON containing `orgId` (`apps/web/app/api/slack/oauth/route.ts`). It was not signed, stored, single-use, or bound to the browser that initiated the flow. The callback verified that the current user was an admin of the org named in the (attacker-controllable) state, but never verified that this browser actually started the authorization.

### Attack chain (pre-fix)
1. Attacker obtains a valid short-lived Slack OAuth `code` for a workspace they control.
2. Attacker forges `state = base64url({orgId: <victim_org>})`.
3. A logged-in victim org admin opens the crafted callback URL.
4. The callback's admin check passes (the victim *is* an admin of their own org), so the attacker workspace is upserted onto the victim org, overwriting any existing integration.
5. A correctly Slack-signed `/octopus` command from the attacker workspace then maps to the victim org and exposes its private tenant context (repos, members + emails, knowledge, PRs, code/review/chat/diagram chunks) to the attacker's Slack.

## Changes
- Issue a high-entropy nonce per OAuth transaction; embed it in an **AES-256-GCM encrypted (tamper-proof) state** alongside `orgId`, `userId`, and an expiry, and set it in an **HttpOnly, SameSite=Lax, Secure** cookie.
- In the callback: decrypt the state, **require the nonce to match the cookie** (binds the flow to the initiating browser; an attacker cannot plant this cookie in a victim's browser), enforce expiry, **require the completing session to be the same user that started it**, and keep the existing admin check as defense in depth.
- Consume the state cookie **single-use**: cleared on every callback response.
- Extract shared constants to `apps/web/lib/slack-oauth.ts`.
- Add regression tests: forged plaintext state, valid state with missing initiator cookie (the CSRF case), cross-user completion, and the genuine happy path.

## Test
```
bun test lib/__tests__/slack-oauth-state.test.ts
# 4 pass, 0 fail
```

## Not included (follow-up)
The report also suggested making `teamId` unique. Investigation against production showed this is **not viable**: one Slack workspace legitimately maps to multiple orgs (different people from the same company connect the same workspace to their own orgs). The correct follow-up is **explicit tenant selection** in the slash-command handler (`commands/route.ts` currently uses `findFirst({ where: { teamId } })` with no ordering, so it resolves ambiguously when one workspace maps to several orgs). Tracked separately.

### DIFF
--- a/apps/web/app/api/slack/callback/route.ts
+++ b/apps/web/app/api/slack/callback/route.ts
@@ -1,8 +1,24 @@
 import { NextRequest, NextResponse } from "next/server";
-import { headers } from "next/headers";
+import { headers, cookies } from "next/headers";
 import { auth } from "@/lib/auth";
 import { prisma } from "@octopus/db";
-import { encryptString } from "@/lib/crypto";
+import { encryptString, decryptJson } from "@/lib/crypto";
+import { SLACK_OAUTH_STATE_COOKIE } from "@/lib/slack-oauth";
+
+type SlackOAuthState = {
+  orgId: string;
+  userId: string;
+  nonce: string;
+  exp: number;
+};
+
+// The state cookie is single-use: clear it on every callback response so a
+// captured code+state pair cannot be replayed against a stale cookie.
+function redirectClearingState(url: URL): NextResponse {
+  const response = NextResponse.redirect(url);
+  response.cookies.delete(SLACK_OAUTH_STATE_COOKIE);
+  return response;
+}
 
 const SLACK_EVENT_TYPES = [
   "review-requested",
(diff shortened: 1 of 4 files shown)


redoh:
Thanks @octopus — addressed the review in dcaea5f:

- 🟡 **Expiry check before I/O** (`callback`): moved the `state.exp` guard to immediately after decryption, ahead of the cookie read, `getSession()`, and the membership query. Stale/replayed-after-expiry state is now rejected before any I/O. ✅ Fixed.
- 💡 **`SlackOAuthState` location**: moved the type into `lib/slack-oauth.ts` next to the shared cookie name and TTL, and imported it in the callback. ✅ Fixed.
- 💡 **Test constant naming**: renamed `ATTACKER_TEAM` → `MOCK_SLACK_TEAM` (and the mock token fields) so the happy-path success assertion no longer reads as asserting an attacker workspace. ✅ Fixed.
- 🔵 **Linear/Jira still use unprotected base64url state**: valid and confirmed, but intentionally **out of scope** for this PR to keep it focused on the Slack flow. Tracking as a separate follow-up to apply the same encrypted-state + cookie-nonce pattern (`encryptJson`/`decryptJson`) to Linear and Jira. 👍

Tests: `bun test lib/__tests__/slack-oauth-state.test.ts` → 4 pass, 0 fail. Ready for re-review.

PR #543, opened 2026-06-26
feat(providers): add Ollama provider (env-configured)

cemoso:
## What
Implement the **Ollama** provider that the repo's `.env.example` already anticipated (`OLLAMA_SERVER_URL` / `OLLAMA_USERNAME` / `OLLAMA_PASSWORD`) but had no code for. Ollama exposes an OpenAI-compatible Chat Completions API, so it reuses the OpenAI SDK with a custom `baseURL`.

- `providers/ollama.ts`: base URL from `OLLAMA_SERVER_URL` (default `http://localhost:11434`), optional HTTP Basic auth (`OLLAMA_USERNAME`/`OLLAMA_PASSWORD`) for a proxied host, strips the `ollama:` model-id namespace prefix
- registered in `providers/index.ts`; `AiProvider` union + `"ollama:"` `PROVIDER_FALLBACK` prefix; `getOrgKeyForProvider` returns `null` (Ollama is env-configured, no per-org key)
- `ai-usage`: `ollama` is free — it runs on the operator's own infra and never bills the platform (`usedOwnKey` true → no credit deduction)

## Why
Continues upstreaming the provider set unblocked by the registry (#540). Ollama is configured per **deployment** via the env vars the repo already documents, so this PR aligns code with those existing keys.

## Scope notes
- **Env-configured / operator-trusted** — there is intentionally **no per-org base-URL column** and therefore **no SSRF surface** in this PR. The `orgId` registry widening and the SSRF URL guard (`url-validation.ts`) arrive with the **ACPX / OpenCode** providers, which take genuinely *org-supplied* gateway endpoints (and will ship a settings writer for them).
- `supportsJsonSchema: false` — Ollama can be asked for JSON but doesn't enforce a schema, so the (currently unwired) `responseSchema` is ignored for it.
- No models are seeded; admins add Ollama models to `AvailableModel` (priced; Ollama is free so credits aren't deducted regardless).

## Test plan
- [x] `bun run typecheck` clean
- [x] `bun run lint` clean
- [x] `cd apps/web && bun test` — 0 fail (400 tests)
- [x] `bun run build` OK
- [x] no schema/migration change (env-only)

### DIFF
--- a/apps/web/lib/ai-router.ts
+++ b/apps/web/lib/ai-router.ts
@@ -21,6 +21,7 @@ const PROVIDER_FALLBACK: Record<string, AiProvider> = {
   // Native OpenRouter ids are vendor/model (e.g. "openai/gpt-4o") and resolve
   // via the AvailableModel DB cache above, not this prefix.
   "openrouter/": "openrouter",
+  "ollama:": "ollama", // namespaced local models, e.g. "ollama:qwen2.5-coder:32b"
 };
 
 let providerCache: Map<string, AiProvider> | null = null;
@@ -98,6 +99,8 @@ function getOrgKeyForProvider(keys: OrgKeys, provider: AiProvider): string | nul
     case "google": return keys.googleApiKey;
     case "grok": return keys.grokApiKey;
     case "openrouter": return keys.openrouterApiKey;
+    // Ollama runs on the operator's own infra — env-configured, no per-org key.
+    case "ollama": return null;
   }
 }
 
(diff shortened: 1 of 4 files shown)


cemoso:

## Changes Made
| # | Finding | Action |
|---|---------|--------|
| 1 | Security: `OLLAMA_SERVER_URL` unsanitized (SSRF) | Added `normalizeServerUrl()` — parses the URL, requires `http(s)`, reduces to a clean origin, and throws on a malformed value. Private/loopback hosts remain allowed because this is a deployment-**operator** env var (not user-supplied), so there is no SSRF attacker surface; full private-host blocking lands with the ACPX/OpenCode providers that take *org-supplied* URLs. |
| 2 | Error Handling: client caches first call; env change needs restart, undocumented | Documented the process-lifetime singleton (env read once at first use; restart to change) — consistent with the platform-key singletons in the other providers. |

No schema/migration change. `typecheck` / `lint` / `bun test` (400, 0 fail) / `build` all green.

PR #546, opened 2026-06-26
feat(seed): make Codex the platform default review model

cemoso:
## What
Make **Codex** the platform default review model. Promote the existing `gpt-5.3-codex` seed row to `isPlatformDefault: true`, and add a seed integrity guard asserting exactly one `llm` row carries the flag.

## Why
`ai-client.ts` resolves the review model as `org.defaultModelId ?? getPlatformDefault("llm") ?? HARDCODED_REVIEW_MODEL`. Today no model is marked default, so every org without a custom model falls back to the hardcoded `claude-sonnet-4-6`. This marks `gpt-5.3-codex` as the platform default — coding-focused and cheaper than Sonnet ($1.75/$14 vs $3/$15 per 1M tokens).

## ⚠️ Product change — please confirm
This **changes the default review model for every org that hasn't set its own** (`claude-sonnet-4-6` → `gpt-5.3-codex`). Orgs with an explicit `defaultModelId` are unaffected. Flagging it explicitly since it shifts behavior platform-wide.

- I used origin's existing curated codex model (`gpt-5.3-codex`). If you'd prefer the cheaper **`codex-mini-latest`** ($1.5/$6) as the default, that's a one-line follow-up (it would also add that model row to the catalog).

## Test plan
- [x] `bun run typecheck` clean (incl. `packages/db` `tsc`)
- [x] seed integrity guard: exactly one `llm` `isPlatformDefault` row (gpt-5.3-codex)
- [x] no schema change (`isPlatformDefault` column already exists)

### DIFF
--- a/apps/web/lib/ai-client.ts
+++ b/apps/web/lib/ai-client.ts
@@ -1,6 +1,8 @@
 import { prisma } from "@octopus/db";
 
-export const HARDCODED_REVIEW_MODEL = "claude-sonnet-4-6";
+// Last-resort fallbacks when the DB has no platform default (e.g. an unseeded
+// instance). Kept in sync with the seed's isPlatformDefault rows.
+export const HARDCODED_REVIEW_MODEL = "gpt-5.3-codex";
 export const HARDCODED_EMBED_MODEL = "text-embedding-3-large";
 
 async function getPlatformDefault(category: "llm" | "embedding"): Promise<string | null> {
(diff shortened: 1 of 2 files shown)


cemoso:

## Changes Made
| # | Finding | Action |
|---|---------|--------|
| 1 | Consistency: `gpt-5.3-codex` drifts from `HARDCODED_REVIEW_MODEL` (`claude-sonnet-4-6`) | Set `HARDCODED_REVIEW_MODEL = "gpt-5.3-codex"` so the last-resort fallback matches the seeded platform default. |
| 2 | Consistency: no embedding platform default | Marked `text-embedding-3-large` (== `HARDCODED_EMBED_MODEL`) as the embedding `isPlatformDefault`, so `getPlatformDefault("embedding")` resolves a row instead of always falling back. |
| 3 | (hardening) | Extended the seed integrity guard to assert exactly one default **per category** (llm + embedding). |

`typecheck` (incl. `packages/db` tsc) / `lint` clean. No schema change.

PR #549, opened 2026-06-28
feat(landing): derive CLI install URL from the current origin

cemoso:
## What
The landing-page CLI install one-liners hardcoded `https://octopus-review.ai`. Derive the base URL from `NEXT_PUBLIC_APP_URL` (SSR/first paint) and `window.location.origin` after mount, so self-hosted / reverse-proxied / custom-domain deployments render the correct `curl`/`irm` host. Falls back to `https://octopus-review.ai` (unchanged hosted behavior); npm commands unchanged.

## Why
Self-hosters copying the landing-page install command currently get the hosted domain instead of their own instance.

## Test plan
- [x] typecheck / lint / build clean
- [x] hydration-safe: initial state = env-derived default (matches SSR), then `window.location.origin` applied post-mount

### DIFF
--- a/apps/web/components/landing-cli-install.tsx
+++ b/apps/web/components/landing-cli-install.tsx
@@ -7,28 +7,38 @@ import { TrackedLink } from "@/components/tracked-link";
 type Platform = "mac-linux" | "windows";
 type Method = "one-liner" | "npm";
 
-const installCommands: Record<Platform, Record<Method, { comment: string; command: string }>> = {
-  "mac-linux": {
-    "one-liner": {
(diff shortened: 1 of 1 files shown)


cemoso:
@octopus Fixed — both NEXT_PUBLIC_APP_URL and window.location.origin now go through `safeBaseUrl()`, which parses via `new URL()` and returns only the clean `scheme://host[:port]` origin (rejecting non-http(s)/unparseable values). No path/query/shell-metacharacters can reach the displayed install command. typecheck/lint/build green. Ready for re-review.

PR #554, opened 2026-06-28
feat(auth): self-hosted email+password login (gated)

cemoso:
## What
Adds **email/password sign-in + sign-up**, a **first-boot admin seed**, and **forced first-login password change** for self-hosted deployments — all gated behind `NEXT_PUBLIC_OCTOPUS_SELF_HOSTED`.

- **`auth.ts`** — the `emailAndPassword` block is included **only when self-hosted**; on the SaaS the key is omitted entirely.
- **`bootstrap-admin.ts`** — seeds `admin@example.com` / `change-me-now` on an empty user table, sets `mustChangePassword`; double-gated (instrumentation `if` + self-check) + idempotent (`DISABLE_ADMIN_SEED` to skip).
- **Login** — password/signup UI behind a `passwordAuth` prop (computed server-side); reconciled onto the SSR login split. Magic-link + OAuth untouched.
- **New pages** — `/forgot-password`, `/reset-password`, `/change-password`; middleware redirects them to `/login` on non-self-host builds.
- **`(app)` layout** — redirects to `/change-password` while `mustChangePassword`; new `User.mustChangePassword` column + migration.
- **`auth-client`** — `signUp`, `requestPasswordReset`, `resetPassword`, `changePassword`.

## SaaS safety (strict no-op when the flag is unset)
The flag is unset on the SaaS build (set only by `release.yml` for the GHCR self-host image). With it unset:
- `auth.ts` omits `emailAndPassword` entirely, so **all** password endpoints self-reject — sign-in → `EMAIL_PASSWORD_DISABLED`, sign-up → `EMAIL_PASSWORD_SIGN_UP_DISABLED`, request-password-reset → `RESET_PASSWORD_DISABLED`, reset-password → `INVALID_TOKEN` (no token can be minted), change-password → `CREDENTIAL_ACCOUNT_NOT_FOUND`.
- bootstrap seed never runs (double-gated; also no-ops on a non-empty table).
- login shows OAuth + magic-link only; `mustChangePassword` is never set, so the layout redirect never fires; the password pages redirect to `/login`.

> Note: omitting the whole block matters — `enabled: false` alone would leave request/reset-password live, because Better Auth gates those on the **presence** of `sendResetPassword`, not on `enabled`.

## Validation
Adversarial pre-PR pass (SaaS-blast-radius / auth-correctness / scope). The blast-radius lens caught the `sendResetPassword`-presence leak above; fixed by omitting the block, then re-verified per-endpoint against the installed Better Auth → **SAAS-SAFE**. Auth-correctness: self-host flow works end-to-end, existing magic-link/OAuth preserved. Scope: clean.

## Test plan
- [x] db:generate, typecheck, lint, build clean

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -119,10 +119,16 @@ NEXT_PUBLIC_APP_URL=http://localhost:3000
 ADMIN_EMAILS=
 # Self-hosted build metadata (baked into the GHCR image by release.yml).
 # NEXT_PUBLIC_OCTOPUS_SELF_HOSTED=true surfaces the in-app Updates page +
-# settings nav entry; NEXT_PUBLIC_APP_VERSION is the running version compared
-# against the latest GitHub release. Leave both unset for the hosted build.
+# settings nav entry, AND enables email/password sign-in + sign-up (off on the
+# hosted SaaS, which uses OAuth + magic-link only). NEXT_PUBLIC_APP_VERSION is
+# the running version compared against the latest GitHub release. Because these
+# gate client UI, set them at BUILD time (build args) — not just runtime — when
+# building from source. Leave both unset for the hosted build.
 NEXT_PUBLIC_APP_VERSION=
 NEXT_PUBLIC_OCTOPUS_SELF_HOSTED=
+# Set true to skip the first-boot admin@example.com seed (self-host only). Use
+# when you provision the first user out-of-band (e.g. via OAuth/magic-link).
+DISABLE_ADMIN_SEED=
 
 # Review worker
 # Only containers with this set to "true" will consume pg-boss "process-review"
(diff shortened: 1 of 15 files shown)


cemoso:
Addressed the Security findings:

1. **Hardcoded default credential** → the first-boot seed password is now **random per install** (`randomBytes` base64url), printed once to the boot log; operators can pin one via `OCTOPUS_ADMIN_PASSWORD` (validated ≥10 chars, fails closed otherwise). Nothing hardcoded ships; the account is still forced through `mustChangePassword` on first sign-in.
2. **Unauthenticated POST abuse** → added explicit Better Auth `rateLimit` customRules on the password endpoints (`/request-password-reset` 3/min, `/reset-password` 5/min, `/sign-in/email` 10/min, `/sign-up/email` 5/hr). Scoped to self-host (inside the `emailAndPassword` block that's omitted on the SaaS), so the SaaS is unchanged. Note: rate-limit counters use Better Auth's default in-memory store, so limits are per-replica on multi-replica self-host deployments.

Re-verified: db:generate / typecheck / lint / build clean; SaaS remains a strict no-op for all password endpoints. @octopus please re-review.

PR #558, opened 2026-06-29
feat(cli): octp — native CLI (onboarding, review, agent serve, doctor)

cemoso:
## What
The self-contained **octp** CLI (`apps/cli`, ~5.4k LOC, ink TUI, Bun-compiled binary). Subcommands: first-run **onboarding wizard**, **`octp review`** (uses the local-review + index-local backends from Slices 1–2), **`octp agent serve`** (the LLM-agent bridge from Slice 3), **`octp doctor`**. Device-flow auth against existing `/api/cli/auth/*`; credentials in `~/.octopus` (mode 0600). **No `@octopus/*` runtime imports** — only the `@octopus/tsconfig` build dep (already upstream).

## CI / monorepo
- `bun.lock` regenerated (adds ink / ink-select-input / ink-text-input / react-devtools-core); `bun install --frozen-lockfile` passes.
- turbo runs the CLI's typecheck (`tsc --noEmit`) + build (`bun build --target node --minify`) — both green. (`bun test` is local-only; 63/0 here.)
- **Not** added to `.dockerignore`: the web Dockerfile uses `--frozen-lockfile` which references the `@octopus/cli` workspace, so excluding it would break the frozen install. The source in the build context is negligible and never enters the final image (only `apps/web/.next/standalone` is copied).

## Constant sync (CLI hand-duplicates a few server values)
- `MAX_DIFF_BYTES` = 500KB == `cli-limits.ts` `MAX_LOCAL_REVIEW_DIFF_BYTES` ✓
- model list mirrors `cost.ts`/seed; provider slugs mirror `ai-router` `PROVIDER_FALLBACK`.

## Scope
- The `octp-v*` release workflow + web-served installers are **Slice 5** (next). `apps/cli/install/` scripts are included (part of the CLI) but resolve a binary only once a release is cut.

## Test plan
- [x] frozen-lockfile / typecheck / lint / build clean; `cd apps/cli && bun test` = 63/0.

### DIFF
--- a/apps/cli/AGENTS.md
+++ b/apps/cli/AGENTS.md
@@ -1,0 +1,44 @@
+# AGENTS.md — apps/cli
+
+Scoped rules for this package. The root [AGENTS.md](../../AGENTS.md) still applies.
+
+This package is **the Octopus CLI** (`octp` binary). It replaces the previous npm-published `@octp/cli` — distributed natively via curl/irm, includes onboarding as a built-in TUI wizard.
+
+## Stack constraints
+
+- Built with **ink ^6** (React for terminals). Do **not** introduce a second TUI library.
+- No Next.js, no Prisma, no database access. This package runs on the user's laptop with zero infrastructure.
+- All persistence is to files under `$OCTOPUS_HOME` (default `~/.octopus/`). HTTP calls allowed for explicit needs only (auth, provider validation, repo listing).
+- Secrets stay in `byok.json` / `credentials` — never in `config.json` (the prefs file must remain safe to `cat`).
+
+## Subcommand conventions
+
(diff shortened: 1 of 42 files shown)


cemoso:
Addressed the review:
- **SHA256 case mismatch** — install.sh now lowercases both the expected (SUMS) and computed hashes before comparing, so an uppercase `SHA256SUMS.txt` entry can't falsely reject a valid binary. (It never falsely *passed* — `!=` only over-rejects.)
- **install.sh temp leak** — `mv` is now guarded; on failure the script exits with the EXIT cleanup trap still active, so `$tmp_file` is removed.
- **OCTOPUS_HOME** — resolved to an absolute path (`path.resolve`) so a relative/`..` value can't land config/credentials relative to the cwd. (It's trusted operator/test config — same-user, same-machine — so this is hardening, not a privilege boundary.)
- **SIGTERM** — already handled in `agent serve` (alongside SIGINT, lines 156-157); SIGPIPE uses the Node default.

@octopus please re-review.

PR #572, opened 2026-06-30
feat(telemetry): vendor cross-org telemetry console (PR D)

cemoso:
## What

**PR D of 4 — the final slice.** A vendor (Octopus staff) super-admin console with a cross-org view of live telemetry, built to a strict privacy contract. Completes the feature.

## Super-admin (vendor staff)
- `lib/superadmin.ts getSuperAdmin()` — identity is an **env allowlist** (`OCTOPUS_SUPERADMIN_EMAILS`) resolved to the session user **by immutable id**, then required to be `emailVerified` **and** allowlisted. There is **no DB-settable flag**, so no in-app path escalates to super-admin. Feature is off when the allowlist is empty.

## Console
- **`/admin/telemetry`** (`force-dynamic`; outside the org-scoped `(app)` group): `notFound()` for non-super-admins **and on self-host** (single-tenant has nothing to aggregate); human access is **audit-logged** on render.
- **`GET /api/admin/telemetry`** — dual auth: `ADMIN_API_SECRET` bearer (machine, matching the other `/api/admin/*` routes) **or** a super-admin session; **404** to anyone else so the route isn't disclosed. **Programmatic (machine) access is audit-logged**; session polls aren't re-logged (the page render already logs each console open) to avoid flooding the audit log.

## Privacy contract
- `lib/vendor-telemetry.ts` — org-level **aggregates by default** (orgs-enabled, online member/agent counts, 24h activity volume); **no member identities**. Member **names** appear per-org **only** when that org set `allowVendorMemberVisibility` — a **separate, owner-only** opt-in, distinct from the org's own internal `liveTelemetryEnabled`. So enabling internal monitoring never exposes named employees to the vendor. Cross-org totals are **scoped to enabled orgs** (never counting orgs that merely run an agent or are soft-deleted). Cached 5s; presence reads parallelized.
- `toggleVendorMemberVisibility` (owner-only, audited) + a settings switch shown only when telemetry is active and not self-hosted.

## Folded-in PR C review notes
- `/api/activity` no longer returns the unused `metadata` field.
- `(app)/layout` cheap-gates the entitlement lookup behind the raw `liveTelemetryEnabled` flag (zero extra query for the common telemetry-off case).

## Notes for review (pre-emptions)
- No migration (`allowVendorMemberVisibility` shipped in PR A).
- The `ADMIN_API_SECRET` check uses `===`, matching the existing `/api/admin/*` routes (consistency).
- Caught in validation: the page must be `force-dynamic` — otherwise the build-time prerender (allowlist empty at build) would bake in a 404 served to everyone.
- Vendor totals were tightened post-review to scope to telemetry-enabled orgs (so the header never reflects non-participating orgs).

## Test plan
- `bun run typecheck` ✓ (3/3) · `eslint` ✓ (0 errors) · `bun run build` ✓ (`/admin/telemetry` + `/api/admin/telemetry` registered, dynamic)
- `bun test` ✓ **457 pass / 0 fail**.

### DIFF
--- a/apps/web/app/(app)/actions.ts
+++ b/apps/web/app/(app)/actions.ts
@@ -964,6 +964,61 @@ export async function toggleTelemetryOptOut(optedOut: boolean): Promise<{ error?
   return {};
 }
 
+/**
+ * Org opt-in to letting Octopus (vendor) staff see member-level detail in the
+ * cross-org console — SEPARATE from liveTelemetryEnabled (internal monitoring).
+ * Owner-only: authorizing a third party to see named members is a higher-trust
+ * consent than enabling internal monitoring. Audited under "admin".
+ */
+export async function toggleVendorMemberVisibility(
+  _prevState: { error?: string; success?: boolean },
+  formData: FormData,
+): Promise<{ error?: string; success?: boolean }> {
+  const user = await getUser();
+  const reqHeaders = await headers();
+  const cookieStore = await cookies();
+  const orgId = cookieStore.get("current_org_id")?.value;
+  if (!orgId) return { error: "No organization selected." };
+
+  const member = await prisma.organizationMember.findFirst({
+    where: { organizationId: orgId, userId: user.id, deletedAt: null },
+    select: { role: true },
+  });
+  if (!member || member.role !== "owner") {
+    return { error: "Only the organization owner can change vendor visibility." };
+  }
+
+  const allowed = formData.get("allowed") === "true";
+  const before = await prisma.organization.findUnique({
+    where: { id: orgId },
+    select: { allowVendorMemberVisibility: true },
+  });
+  if (before && before.allowVendorMemberVisibility === allowed) return { success: true };
+
+  await prisma.organization.update({
+    where: { id: orgId },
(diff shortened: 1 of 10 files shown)


cemoso:

## Changes since last review (commit aaa3a48)

| # | Finding | Action |
|---|---------|--------|
| 1 | 🟠 Security 3/5 — module-level cache holds cross-org data (serverless cross-request risk) | **Fixed** — removed the process-global cache in `lib/vendor-telemetry.ts`; recompute per call (queries are grouped + parallelized; the vendor console is low-traffic). Eliminates any chance of cross-request reuse of cross-tenant data. |
| 2 | 🟡 Performance — N+1 user lookups in the vendor loop | **Fixed** — member display names now resolved in a single `prisma.user.findMany` across all member-visible orgs (was one query per org). |
| 3 | 🔵 Error handling — silent swallow in poll callbacks | Intentional + commented (transient fetch errors retry on the next tick). |

Authorization, the aggregates-by-default privacy contract, and the separate owner-only `allowVendorMemberVisibility` gate are unchanged.

PR #581, opened 2026-07-02
ci(release): also build the hosted-deploy image variant

cemoso:
## What
Adds a second `docker/build-push-action` step to `release.yml` building the **hosted-deploy variant** of the same commit: `selfHosted` off + `NEXT_PUBLIC_GITHUB_APP_SLUG`/`NEXT_PUBLIC_PUBBY_KEY` build args (client-exposed identifiers, supplied as repo **variables** — already set). Pushes immutable `X.Y.Z-prod` + the `latest-web`/`latest-engine` aliases the hosted deployment pulls. Adds OCI `revision`/`version` labels to both variants.

## Why
The existing image bakes `NEXT_PUBLIC_OCTOPUS_SELF_HOSTED=true` (correct for the self-host distribution), so the hosted deployment could never use it — it has been building on the production host at deploy time: slow (~8 min on-box Next build competing with live traffic), and it ships a working tree rather than a reviewed commit. With this, deploys pull a CI-built, SHA-labeled image instead.

## Notes for review
- The prod step is `if: vars.NEXT_PUBLIC_PUBBY_KEY != ''` so forks/self-host clones without the variables skip it cleanly.
- Pushing the `latest-web`/`latest-engine` aliases does **not** restart anything — running services only re-resolve tags on an explicit service update.
- Both steps share the GHA layer cache; only the Next build stage re-runs for the second variant (build args differ).
- `NEXT_PUBLIC_*` values are inlined into client bundles by definition — they are identifiers, not secrets, hence repo variables.

## Test plan
Covered by cutting the next release tag: expect both variants published (`X.Y.Z`, `latest`, `X.Y.Z-prod`, `latest-web`, `latest-engine`) with OCI labels; a `--skip-build` deployment then pulls `latest-web` instead of building on the box.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -74,8 +74,10 @@ jobs:
             # This GHCR image is the self-hosted distribution, so the Updates
             # page + nav entry (gated on this build-time flag) ship enabled.
             NEXT_PUBLIC_OCTOPUS_SELF_HOSTED=true
-          cache-from: type=gha
-          cache-to: type=gha,mode=max
+          # Distinct cache scope per variant — with the shared default scope the
+          # two builds would overwrite each other's cache every release.
+          cache-from: type=gha,scope=selfhost
+          cache-to: type=gha,mode=max,scope=selfhost
 
       # Hosted-deployment variant of the same commit. NEXT_PUBLIC_* values are
       # inlined into the client bundle at build time, so the hosted instance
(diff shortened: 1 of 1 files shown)


cemoso:

| # | Finding | Action |
|---|---------|--------|
| 1 | Performance 3/5 — shared GHA cache scope | Fixed: distinct `scope=selfhost` / `scope=prod` on cache-from/cache-to of each build step. |
| 2 | Code Quality — guard checks one var | Fixed: `if:` now requires both `NEXT_PUBLIC_PUBBY_KEY` and `NEXT_PUBLIC_GITHUB_APP_SLUG`. |

PR #593, opened 2026-07-04
feat: per-org provider config — ollama base URL, ACP + OpenCode gateway credentials

cemoso:
Adds nullable Organization columns (ollamaBaseUrl, acpBaseUrl/acpApiKey, opencodeBaseUrl/opencodeApiKey) with an expand-only migration; per-org values override env defaults in the ollama/acp/opencode providers; gateway keys are encrypted at rest exactly like existing BYOK keys (encryptString on write, decryptStringMaybeLegacy on read); base URLs are SSRF-validated via the existing validateProviderUrl; updateApiKeys accepts the new fields. Also removes the per-provider client cache in openai-gateway in favor of per-call clients (required for per-org URLs; construction cost is negligible vs the LLM call). (cemoso/octopus#19, #44, #45.)

### DIFF
--- a/apps/web/app/(app)/actions.ts
+++ b/apps/web/app/(app)/actions.ts
@@ -250,6 +250,12 @@ export async function updateApiKeys(
   const cohereApiKey = (formData.get("cohereApiKey") as string)?.trim() || null;
   const grokApiKey = (formData.get("grokApiKey") as string)?.trim() || null;
   const openrouterApiKey = (formData.get("openrouterApiKey") as string)?.trim() || null;
+  // Per-org provider config: gateway/base-URL overrides + gateway API keys.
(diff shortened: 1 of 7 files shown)


cemoso:

| Finding | Action |
|---|---|
| Add save-time URL validation | Fixed: updateApiKeys validates the three base URLs via validateProviderUrl before persisting; standard error shape on rejection. |
| URLs cannot be cleared once set | Fixed: presence-aware handling for the 5 new fields — absent=unchanged, empty=clear (null), non-empty=validate/encrypt+set. (Existing BYOK keys clear via removeApiKey; noted, untouched.) |
| Env gateway URLs may fail SSRF checks | Fixed: SSRF policy applies only to per-org (user-supplied) URLs; env-configured URLs validated with hosted:false (http(s)+origin normalization, private hosts allowed — operator-controlled). |

cemoso:
@octopus review please

PR #609, opened 2026-07-05
feat(blog-api): scoped service tokens for external apps

cemoso:
## What

Turns the single platform-global blog token into a **scoped `ServiceToken`** so external apps (Claude routines, MCP integrations, other tooling) can be granted **least-privilege** access instead of one all-or-nothing key.

Scopes shipped: `blog:read`, `blog:create`, `blog:update`. (`blog:delete` + the DELETE endpoint are deferred by design.)

## Changes

- **Schema:** `BlogApiToken` → `ServiceToken` — the model is renamed but keeps `@@map("blog_api_tokens")`, so the rename is **zero-DDL**. The migration only **adds** `scopes` / `expiresAt` / `createdBy` columns and **backfills** every existing token with `blog:read,blog:create,blog:update` so current routines keep working uninterrupted. Deny-by-default: an empty scope set is inert (expand-only, migrate-check safe).
- **`lib/scopes.ts`:** `resource:action` registry, `normalizeScopes` (trim/lowercase/dedupe, reject unknown or empty), `hasScopes` (deny-by-default, exact membership, **no wildcard**). Unit-tested in `lib/__tests__/scopes.test.ts` (8 tests).
- **`/api/blog`:** auth swapped to `ServiceToken`. `GET` requires `blog:read`, `POST` `blog:create`, `PATCH` `blog:update` — **401** when unauthenticated vs **403** when the token lacks the scope. `PATCH` expanded to edit `title`/`content`/`excerpt`/`coverImageUrl`/`status` (+ publish) alongside the existing `audioUrl`/`tags`/`category`; reading time recomputed on content edit; slug immutable. `audioUrl` validation is **fail-closed**; new `coverImageUrl` host allowlist (R2 host + `cdn.octopus-review.ai`). Audit log line records the **token prefix only** (never the secret) on create/update.
- **`/admin/tokens`:** super-admin-only mint UI (reveal-once display + copy) and revoke. Server actions **re-check super-admin** on every call (defense in depth). Route + actions 404 / no-op on self-host.
- **`scripts/mint-service-token.ts`:** universal mint path for self-host / CI, where the admin UI isn't available.

## Why

The blog token was platform-global and unscoped — handing it to an external automation granted full create/edit. Scoped tokens let each app hold only what it needs, and the reveal-once + revoke flow gives a clean rotation story.

## Test plan

- `bun run typecheck` — clean
- `bun run lint` — 0 errors
- `bun run build` — succeeds; `/admin/tokens` and `/api/blog` both register
- `cd apps/web && bun test lib/__tests__/scopes.test.ts` — 8/0
- Migration is expand-only (ADD COLUMN + backfill UPDATE), passes migrate-check

## Notes for review

- The `ServiceToken` rename preserves `@@map`, so **no table rename DDL** is emitted — verified by the migration containing only `ADD COLUMN` + backfill.
- `findUnique({ where: { tokenHash, deletedAt: null } })` uses Prisma's extended-unique filter so a revoked token resolves to null.
- Structured deny-by-default: `hasScopes` returns false for null/empty/partial/wildcard inputs (covered by tests).

### DIFF
--- a/apps/web/app/admin/tokens/actions.ts
+++ b/apps/web/app/admin/tokens/actions.ts
@@ -1,0 +1,54 @@
+"use server";
+
+import { prisma } from "@octopus/db";
+import { revalidatePath } from "next/cache";
+import { getSuperAdmin } from "@/lib/superadmin";
+import { generateServiceToken, hashToken, serviceTokenPrefix } from "@/lib/api-auth";
+import { normalizeScopes } from "@/lib/scopes";
+
+// Super-admin is re-checked inside EVERY action (defense in depth — the page
+// gate alone is not sufficient for state-changing operations).
+
+export async function createServiceToken(
+  formData: FormData,
+): Promise<{ ok: true; token: string } | { ok: false; error: string }> {
+  const sa = await getSuperAdmin();
+  if (!sa) return { ok: false, error: "Forbidden" };
+
+  const name = String(formData.get("name") ?? "").trim();
+  if (!name) return { ok: false, error: "A name is required" };
+
+  const selected = formData.getAll("scopes").map((s) => String(s));
+  let scopes: string[];
+  try {
+    scopes = normalizeScopes(selected);
+  } catch (e) {
+    return { ok: false, error: (e as Error).message };
+  }
+
+  const token = generateServiceToken();
+  await prisma.serviceToken.create({
+    data: {
+      name,
+      tokenHash: hashToken(token),
+      tokenPrefix: serviceTokenPrefix(token),
+      scopes,
+      createdBy: sa.id,
+    },
+  });
+
+  revalidatePath("/admin/tokens");
+  // The plaintext token is returned ONCE for display and never persisted/logged.
+  return { ok: true, token };
+}
+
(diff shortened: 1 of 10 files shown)


cemoso:

## Changes Made
| # | Finding | Action |
|---|---------|--------|
| 1 | Error Handling: `revokeServiceToken` throws on stale id | Fixed — switched to `updateMany({ where: { id, deletedAt: null } })`; a stale/already-revoked id is now a no-op returning `ok:false` instead of throwing P2025 |
| 2 | Code Quality: `hasScopes` vacuous-truth when `required` is empty | Fixed — added an explicit `required.length === 0 → false` guard (deny-by-default) plus a regression test |
| 3 | Security: audit-log token mint/revoke | Fixed — `createServiceToken`/`revokeServiceToken` now log actor id + token prefix + scopes (never the plaintext), matching the `/api/blog` audit lines |

Note on test coverage: the scope library (`normalizeScopes`/`hasScopes`) is unit-tested; route-level enforcement isn't (the `/api/blog` handlers call Prisma directly and there's no route-test/DB-mock harness in this app today), so I left that as a follow-up rather than introduce new test infra in this PR.

PR #613, opened 2026-07-06
feat(admin-api): service-token + blog-audio endpoints; move admin UI to config console

cemoso:
## What

Service-token minting and blog read-aloud management belong in the **`octopus-configuration`** console (the vendor admin surface, octps.oros.boo) — not as `getSuperAdmin()`-gated pages in the public web app, where they 404'd for the vendor and split admin across two apps. This moves the **logic** behind `ADMIN_API_SECRET` endpoints the config app already knows how to call (`lib/octopus-api.ts`), and removes the orphaned pages. The config-app **UI** is a follow-up PR in that repo.

## Changes

- **New `/api/admin/*` endpoints** (ADMIN_API_SECRET bearer, same pattern as `/api/admin/seed-docs` & `/api/admin/telemetry`):
  - `GET /api/admin/service-tokens` — list active tokens + scope registry.
  - `POST /api/admin/service-tokens` — mint (reveal-once plaintext).
  - `DELETE /api/admin/service-tokens/[id]` — revoke (soft-delete, no-op if already gone).
  - `GET /api/admin/blog-audio` — config flags, selected voice, voices, posts.
  - `POST /api/admin/blog-audio` — `save-voice` | `generate` | `clear`.
  - Reuses `lib/scopes` + `lib/api-auth` (so token hash/scopes can't drift from `/api/blog` auth) and `lib/elevenlabs` + `lib/r2` (generate cache-busts the `audioUrl`).
- **Removed** the orphaned `/admin/tokens` and `/admin/blog-audio` pages.
- **Kept** (still used at runtime): `lib/elevenlabs`, `lib/scopes`, `lib/api-auth`, the `/api/blog` scope enforcement, `scripts/mint-service-token.ts`, and `SystemConfig.blogAudioVoiceId`.

## Deploy note

`ELEVENLABS_API_KEY` still needs to be in the **apps/web** app env for blog-audio generation (the endpoint runs here). No new env for the config app beyond its existing `ADMIN_API_SECRET`.

## Test plan

- `bun run typecheck` — clean · `bun run lint` — 0 errors · `bun run build` — succeeds; the 3 endpoints register and the old pages are gone.


cemoso:

## Changes Made
| # | Finding | Action |
|---|---------|--------|
| 1 | 🟠 Public blog revalidation dropped (clear didn't select slug) | Fixed — generate + clear now `revalidatePath("/blog")` + `` `/blog/${slug}` ``; clear selects the slug |
| 2 | Security: non-constant-time compare | Fixed — new `lib/admin-auth.ts` `isAdminApiAuthorized` uses `crypto.timingSafeEqual` (fails closed when the secret is unset) |
| 3 | Consistency: duplicated auth check | Fixed — all three new `/api/admin/*` routes now use the shared `isAdminApiAuthorized` helper |
| 4 | Security: audit logging dropped | Fixed — `save-voice` and `clear` now log (mint/revoke/generate already did) |

On "reuse `lib/api-auth.ts`": that module is hashed **per-org/service** token auth (a different auth type). These endpoints use the **machine `ADMIN_API_SECRET`** bearer — the established pattern for `/api/admin/*` (seed-docs, telemetry, reindex) — so I factored the shared, constant-time check into `lib/admin-auth.ts` rather than conflate the two. typecheck clean, lint 0 errors.

PR #615, opened 2026-07-11
feat(admin): incident comms — failed-review emails + goodwill credits

cemoso:
Operator tooling for provider outages (motivated by today's OpenAI embedding-quota 429s, which failed customer reviews with no way to find or compensate the affected orgs beyond hand-written SQL).

## What this adds

**`octp admin incidents --since 3h [--match 429]`** — lists every org with failed reviews in the window: counts, repos, distinct errors, and the owner/admin emails a notification would reach. Backed by `GET /api/admin/incidents/failed-reviews` (data already existed: `PullRequest.status="failed"` + `errorMessage`).

**`octp admin incidents notify --incident-key openai-429-2026-07-11 --since 3h --credits 5 --send`** — emails each affected org's owners/admins from the new `incident-resolved` DB template and grants goodwill free credits. Backed by `POST /api/admin/incidents/notify`.

**`octp admin credits grant --org <slug> --amount 10 --reason "…"`** — one-off goodwill grant (`POST /api/admin/credits/grant`). First caller of the previously-unused `addFreeCredits()`.

## Safety rails

- **Dry-run by default** at both layers: the endpoint requires `dryRun:false` explicitly; the CLI always prints the dry-run plan first and only executes with `--send` plus a typed `SEND` confirmation (or `--yes`).
- **Idempotent per (incidentKey, org)** via the new `IncidentComm` unique constraint — claim-first before any send; re-runs and concurrent runs can never re-email or re-grant an org. The claim is released only when a send fails *before* emails are recorded, so retries are safe (failure ordering documented in `lib/incidents.ts`).
- **Caps**: 30-day window ceiling on `since` (a `--since 300d` typo must not become a mass mailing) and a $50/org credit cap, both overridable only with an explicit force flag.
- Deleted/banned orgs excluded from the query; orgs with no owner/admin recipients skipped; every send and grant audit-logged (`admin.incident.notify`, `admin.credits.grant`).
- Admin auth follows the existing `ADMIN_API_SECRET` bearer pattern (CLI: `OCTOPUS_ADMIN_SECRET`) — deliberately not the org-scoped device-flow token.

## Notes for review

- `getAdminRecipients` moved from `email.observer.ts` (private) to `lib/org-recipients.ts` — now shared with the incidents engine; observer behavior unchanged.
- `prisma/migrations` is gitignored ("managed manually") but `.gitignore:42` re-includes `migration.sql`, and the baseline is tracked — the new migration follows the baseline's convention (force-added).
- Full web suite: 406 pass / 1 fail — `providerJsonSchema … forbidden keywords` fails identically on clean master (pre-existing, unrelated).

## Verification

10 new unit tests (`lib/__tests__/incidents.test.ts`) for the window parser, credit-note builder, and incident-key validation; `tsc` clean outside stale `.next` artifacts; CLI built and smoke-tested (`admin --help`, missing-secret guard).


cemoso:
Closing in favor of #616 — this PR was accidentally based on a stale local master, so its diff bundled ~160 unrelated commits (the 3/5 review above evaluated that bundle, not the feature). #616 is the same single feature commit rebased onto current master.

Issue #620, opened 2026-07-17
Make PLATFORM_MARKUP configurable + persist charged cost per AiUsage row


cemoso:
## Problem

1. `PLATFORM_MARKUP` is a hardcoded constant in `apps/web/lib/cost.ts` — changing pricing means a code change and redeploy.
2. `AiUsage` stores token counts but not the cost actually charged; costs are recomputed from *current* `AvailableModel` prices, so any price edit silently rewrites history and real margin can never be measured against provider invoices.

## Proposal

- Read the markup from an env var (`PLATFORM_MARKUP`, default 1.2) so pricing is tunable per deployment.
- Add `chargedCostUsd` (cost snapshot at charge time) and `durationMs` columns to `AiUsage`, written at usage-record time. Historical reporting then uses the snapshot; the recompute path stays for rows predating the column.

## Why

Prerequisite for any pricing tuning (#619): you can't adjust a margin you can't measure.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #622, opened 2026-07-17
fix(billing): explicit welcome-credit grant + env-configurable markup

cemoso:
## What

Two small billing-config hardening changes:

1. **Welcome credits granted explicitly.** The first-org creation paths (`lib/org-create.ts`, `app/(app)/actions.ts`) now set `freeCreditBalance: 150` alongside the existing `free_credit` ledger row, and the `organizations.freeCreditBalance` column default drops from 150 to 0 (migration `20260717200000_default_free_credits_zero`). Balance changes now always originate from an explicit grant with a matching ledger entry — the community-org path already worked this way.

2. **`PLATFORM_MARKUP` configurable via env** (#620, first half). Moves the constant out of `calcCost` to a module-level export read from `process.env.PLATFORM_MARKUP`, guarded to finite ≥ 1 with fallback 1.2 so a bad value can never zero out or invert billing. Documented in `.env.example`.

## Verification

- `bun test` in apps/web: all billing tests pass (2 pre-existing nodemailer transport failures are unrelated — they attempt a real SMTP connection).
- `tsc --noEmit` clean.
- Migration is a bare `ALTER COLUMN ... SET DEFAULT`; no data change, reversible.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -113,6 +113,9 @@ STRIPE_SECRET_KEY=
 STRIPE_WEBHOOK_SECRET=
 NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY=
 
+# Billing: multiplier over provider list price when charging credits (default 1.2)
+PLATFORM_MARKUP=1.2
+
 # Cohere (Reranking)
 COHERE_API_KEY=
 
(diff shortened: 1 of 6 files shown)


cemoso:
Review feedback addressed in 6c57c04:

- **Welcome amount duplicated** — extracted `WELCOME_FREE_CREDITS` into `lib/constants.ts`; both creation paths (`lib/org-create.ts`, `app/(app)/actions.ts`) now use it for the balance, ledger amount, `balanceAfter`, and description.
- **Silent markup fallback** — a set-but-invalid `PLATFORM_MARKUP` now logs a warning naming the raw value before falling back to 1.2.
- **Call-site audit** — there are exactly two `organization.create` sites in the app (the two updated); the community-org path already sets `freeCreditBalance: 0` explicitly. No other path grants credits.
- **Migration visibility** — the default-change migration is in this PR: `packages/db/prisma/migrations/20260717200000_default_free_credits_zero/migration.sql`.

`tsc --noEmit` clean; billing tests 15/15. @octopus please re-review.

PR #623, opened 2026-07-17
feat(billing): monthly subscription plans with saved-card renewals

cemoso:
## What

Monthly **Pro ($49/mo → $54 credits)** and **Team ($99/mo → $115 credits)** subscription plans on top of the existing credit system (#619 Tier 1). Credits granted are non-expiring and land in the unified balance; the bonus makes subscribing strictly better than one-off top-ups.

## How — deliberately NOT Stripe Billing

Each period is an **off-session PaymentIntent against the saved card** — the exact machinery auto-reload already uses — swept by a daily `subscription-renewals` pg-boss cron (cloud-only). A real Stripe catalog (invoices/proration/tax) waits until MRR justifies it.

- **Schema**: `planTier`/`planRenewsAt`/`planCancelAtPeriodEnd` on Organization (migration `20260717210000`).
- **Idempotency**: grants keyed on the PaymentIntent/Checkout-session id via the ledger's unique `stripeSessionId` — cron re-runs and webhook redeliveries can never double-grant, and a redelivery never re-stamps the billing anchor.
- **Renewals**: advance from the due date (no anchor drift), calendar-month clamped (Jan 31 → Feb 28). Failed cards retry daily for 7 days, then downgrade to free (granted credits are kept).
- **Card saving**: checkout sessions now set `setup_future_usage: off_session` — also fixes the pre-existing gap where auto-reload could find no saved card. Subscribers without a card start via Checkout; the webhook (`subscription_start`) grants and stamps the plan.
- **Cancel**: cancel-at-period-end / resume, tier persists until the period lapses.
- **Entitlements**: `subscription` counts as a paid txn type; `planTier` exposed for future tier-specific gates.
- **UI**: Subscription card in billing settings (subscribe / switch / cancel / resume), subscription badge in the ledger.

## Verification

- `bun test`: 21/21 billing tests pass, including 5 new renewal tests (grant+advance, cancel-at-period-end, grace retry, post-grace downgrade, duplicate-grant idempotency). The 2 pre-existing nodemailer SMTP failures are unrelated.
- `tsc --noEmit` clean; `prisma generate` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #504, opened 2026-06-09
Consolidate manual_server_sync_*.sql hot-fixes into proper Prisma migrations


redoh:
## Problem

`packages/db/prisma/migrations/` contains six `manual_server_sync_*.sql` files (2026-05-01 through 2026-06-02) alongside regular Prisma migrations. These represent schema changes applied to production by hand, outside the Prisma migration flow.

Consequences:

- Environments can no longer be assumed schema-identical.
- `prisma migrate deploy` replay on a fresh database is unverified and may not reproduce production schema.
- Each new manual patch increases drift risk.

## Proposal

- Fold the manual SQL into proper, ordered Prisma migrations (or a baseline migration) so a clean `migrate deploy` reproduces production exactly.
- Verify by spinning up a fresh database, running migrations, and diffing the schema against production.
- Document the rule: production schema changes go through Prisma migrations only.

PR #624, opened 2026-07-17
fix(queue): create subscription-renewals queue before schedule/work

cemoso:
The 1.0.43 rollout crash-looped review-engine: #623 added a `subscription-renewals` cron (schedule in instrumentation.ts, worker in queue-workers.ts) but never created the queue in `lib/queue.ts` — and pg-boss v12 requires `createQueue` before `schedule()`/`work()`. Swarm health-gated and auto-rolled review-engine back; web converged.

One-block fix following the existing pattern for every other cron queue in `ensureQueues`: `retryLimit: 1`, 30-min expiry (one off-session charge per due org; the daily sweep is idempotent, so a lost run self-heals next day).

`tsc --noEmit` clean. Fix-forward for the v1.0.44 tag.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,9 +67,6 @@ CLAUDE.md
 # root migrations
 /migrations/
 
-# prisma migrations (managed manually)
-packages/db/prisma/migrations/
-
 # openclaw
 .openclaw/
 
--- a/apps/web/lib/queue.ts
+++ b/apps/web/lib/queue.ts
@@ -123,6 +123,16 @@ export async function startQueue(): Promise<PgBoss> {
     expireInSeconds: 300, // 5 min — a single GitHub fetch + upsert
   }).catch(() => {});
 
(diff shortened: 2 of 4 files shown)


cemoso:
Scope expanded — the 1.0.43 rollout surfaced a second, bigger root cause: `.gitignore` has ignored `packages/db/prisma/migrations/` since April ("managed manually"), so the #622/#623 migration files were silently dropped from their commits. The v1.0.43 image ships code reading `Organization.planTier` against a DB that never got the column → `/settings/billing` 500s in prod right now.

This PR now carries the full fix-forward set:
1. `subscription-renewals` queue creation (original scope — un-crashes review-engine)
2. The two migration files, force-added
3. Removal of the migrations gitignore rule — the deploy pipeline applies migrations from the release-tag checkout, so untracked migrations are guaranteed drift (same footgun as #504 / the incident_comms hotfix)

@octopus please review — this is a prod-regression fix-forward, will tag v1.0.44 on merge.

PR #625, opened 2026-07-17
feat(billing): add cards in-app via embedded Stripe Payment Element

cemoso:
## What

Save a card **without leaving the app**. The Payment Methods section now has an **Add card** button that opens a dialog with Stripe's embedded Payment Element (card form in a Stripe-hosted iframe — card data never touches our servers) on a SetupIntent. On confirm, the card becomes the customer's default for off-session charges. No redirect to Stripe.

Subscribing with no saved card now opens this dialog instead of bouncing to Checkout.

## Also fixes a latent off-session bug

Cards saved via Checkout's `setup_future_usage` are *attached* to the customer but not made the *default*, and an off-session PaymentIntent confirmed without an explicit `payment_method` does **not** fall back to attached cards. So auto-reload and subscription renewals would have silently no-op'd. Added `getOffSessionPaymentMethodId` (customer default, else most-recently-attached card) and now pass `payment_method` explicitly on every off-session charge; both paths no-op cleanly only when there is genuinely no card.

## Security

`finalizeCardSetup` verifies the SetupIntent (id comes from the browser) belongs to the calling org's Stripe customer before making it default; both new actions are owner/admin gated via the existing `getOwnerOrgId`.

## Verification

- `bun test`: 25/25 billing tests (added a no-card renewal no-op test); 2 unrelated pre-existing nodemailer failures.
- `tsc --noEmit` clean.
- Note: the card form needs `NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY` (already set in prod).

### DIFF
--- a/apps/web/app/(app)/settings/billing/actions.ts
+++ b/apps/web/app/(app)/settings/billing/actions.ts
@@ -119,6 +119,25 @@ export async function finalizeCardSetup(
       invoice_settings: { default_payment_method: intent.payment_method },
     });
 
+    // One card per customer: detach every previously-attached card so a
+    // "Replace card" leaves exactly the new one (matches the UI copy, and
+    // stops stale cards being reachable for off-session charges). Best-effort
+    // — the new default is already set, so a detach failure must not fail the
+    // save.
+    try {
+      const existing = await getStripe().paymentMethods.list({
+        customer: org.stripeCustomerId,
+        type: "card",
+      });
+      await Promise.all(
+        existing.data
+          .filter((pm) => pm.id !== intent.payment_method)
+          .map((pm) => getStripe().paymentMethods.detach(pm.id)),
+      );
+    } catch (err) {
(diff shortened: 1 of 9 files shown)


cemoso:
All findings addressed in 52483d4:

- **Replace card left old PM attached** — `finalizeCardSetup` now detaches every other attached card after setting the new default, so the customer keeps exactly one (best-effort; the default is already set first).
- **Helper swallowed API errors as "no card"** — `getOffSessionPaymentMethodId` returns null ONLY for a genuinely cardless customer and now propagates Stripe API errors; callers moved the call inside their try so a transient outage counts as a failed renewal (grace retry), never a downgrade. Added a test for exactly this.
- **Unhandled rejection in dialog effect** — the async init IIFE now `.catch`es and surfaces a friendly error.
- **Semicolon** on the closing `describe`.

26/26 billing tests, `tsc --noEmit` clean. @octopus please re-review.

Issue #506, opened 2026-06-09
Harden spend-limit and credit deduction under concurrency


redoh:
## Problem

Two related gaps in billing enforcement under concurrent load:

1. **Spend limit is checked once, at review start** (`apps/web/lib/reviewer.ts` around the `isOrgOverSpendLimit()` call). N concurrent reviews can each pass the check and collectively blow past an org's monthly cap.
2. **Balance can go negative under concurrency.** `deductCredits()` uses a Postgres `FOR UPDATE` lock (`apps/web/lib/credits.ts`), which serializes the write but does not prevent multiple in-flight reviews from being admitted against the same balance. A declined auto-reload card only logs the error and the review continues.

## Proposal

- Re-check spend limit (or reserve estimated cost) at deduction time, not only at admission.
- Consider a reservation model: hold an estimated amount when the review is admitted, settle the difference on completion.
- Define explicit policy for negative balances and failed auto-reload (block new reviews? notify org owner?).

Related: #378 (duplicate credit-low emails under concurrency, closed).

Issue #501, opened 2026-06-09
Add unit tests for lib/credits.ts and the Stripe webhook


redoh:
## Problem

The billing path has zero automated tests:

- `apps/web/lib/credits.ts` (~267 lines): credit deduction with `FOR UPDATE` locking, free-vs-purchased balance priority, adaptive low-balance notification, auto-reload trigger.
- `apps/web/lib/stripe.ts` and `apps/web/app/api/stripe/webhook/route.ts`: payment webhook handling, receipt lookup (which currently swallows all errors silently), credit top-ups.

A regression here directly affects customer money and would currently reach production undetected.

## Proposal

Add unit tests covering at least:

- Deduction priority (free credits before purchased) and exact balance math, including cache-token discounted costs.
- Behavior at zero/insufficient balance and concurrent deduction scenarios.
- Auto-reload trigger conditions and the 5-minute duplicate guard.
- Stripe webhook event handling (mocked Stripe client): successful payment, failed payment, unknown event types.

Related: #451 (run tests in CI), #452 (coverage reporting).

PR #627, opened 2026-07-18
feat(billing): notify org owner when auto-reload payment fails (#506)

cemoso:
## What

Closes the actionable half of #506: a declined auto-reload used to be a silent `console.error`, so an org owner had no idea their card failed until reviews stopped. Now the decline emits an `auto-reload-failed` event and the email observer sends the owner an actionable *"update your card"* email.

## How

- New `auto-reload-failed` event (types.ts) emitted from `triggerAutoReloadIfNeeded`'s catch, carrying the reload amount, current balance, and Stripe decline code.
- Email observer handler with the **same 24h Serializable cooldown-claim** as `credit-low` — a burst of concurrently-failing reloads can't spam the owner (one email per window).
- New `auto-reload-failed` email template seed.
- Billing alert, so it's excluded from the activity feed (parity with `credit-low`).

## Post-paid model, now tested

Two tests lock in behavior the audit flagged:
- **Usage exceeding balance records a truthful negative balance**, not clamped to zero — the tokens are already spent; the ledger must reflect the real overage.
- **Declined reload emits `auto-reload-failed`** with the decline code.

## Deliberately not in scope

The full **reservation model** from #506 (hold estimated cost at admission, settle on completion) is a review-lifecycle change with its own orphaned-hold failure modes on crash. Given metering is post-paid, that's a design decision — not built speculatively on the money path. This PR ships the concrete, safe win; the reservation piece stays open on #506 for a separate design pass. #501's test asks are already covered by `billing.test.ts` (now 28 cases).

## Verification

- `bun test`: 28/28 billing tests; 2 unrelated pre-existing nodemailer failures.
- `tsc --noEmit` clean.

### DIFF
--- a/apps/web/lib/__tests__/billing.test.ts
+++ b/apps/web/lib/__tests__/billing.test.ts
@@ -130,6 +130,17 @@ mock.module("@octopus/db", () => ({
 
 let mockOffSessionPaymentMethodId = mock(() => Promise.resolve("pm_default" as string | null));
 
+// Capture emitted events; mocking the bus also stops the real observers (which
+// hit prisma methods this harness doesn't stub) from registering on import.
+let emittedEvents: Array<{ type: string; [k: string]: unknown }> = [];
+mock.module("@/lib/events/bus", () => ({
+  eventBus: {
+    emit: (e: { type: string }) => emittedEvents.push(e as never),
+    on: () => {},
+    off: () => {},
+  },
+}));
+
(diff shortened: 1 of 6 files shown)


cemoso:
Addressed the code-narrowing finding in the latest commit — `reason` is now set only when `err.code` is actually a string, so it can never become the literal "undefined". 28/28 tests, tsc clean. @octopus please re-review.

Issue #619, opened 2026-07-17
Epic: pricing tier ladder on top of the credit system


cemoso:
## Goal

Evolve monetization beyond one-off credit top-ups, reusing what's already built. Credits stay the base unit — no per-seat pricing (contradicts our published positioning).

## Ladder (in build-effort order)

1. **Volume credit bonuses** — larger top-ups grant bonus credits (e.g. $100 → 110 credits). Pure config in the purchase dialog + existing `addCredits`/`addFreeCredits`; no new Stripe objects.
2. **Recurring Pro/Team membership without Stripe Billing** — model a monthly plan as a scheduled off-session charge reusing the existing auto-reload machinery (`lib/credits.ts`), plus a `planTier` flag in `lib/entitlements.ts` to gate features. Gives recurring revenue + a feature gate with minimal new code; migrate to a real Stripe catalog (invoices/proration/tax) only once the tier shows traction.
3. **Tier the feature set** — anchor the paid tier on merge gating (#618) and invoice/reporting features (#481, #486, #480, #482); revisit live-telemetry entitlement so the tier ladder has real differentiation.
4. **Enterprise / BYOK platform fee** — orgs running on their own provider keys (per-org provider config, PRs #593/#594) consume platform infrastructure without credit spend; introduce a flat platform fee for cloud-hosted BYOK orgs.

## Dependencies

Billing-path hardening first: #506 (spend-limit concurrency) and #501 (credits/webhook tests) before scaling paid usage.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #628, opened 2026-07-18
feat(billing): volume bonus on credit top-ups (#619 Tier 0)

cemoso:
## What

The smallest revenue lever from the tier-ladder epic (#619): larger one-off top-ups grant bonus credits. **$100+ → 5%, $500+ → 10%.** Rates deliberately stay below the subscription per-dollar bonus (Pro 10% / Team ~16%) so subscribing remains the better deal.

## How

- `volumeBonusUsd(amountUsd)` — pure tiered function in `plans.ts`, single source of truth shared by the dialog and the webhook (highest matching tier wins, rounded to cents, guards non-positive input).
- **Bonus is a separate `free_credit` ledger row**, not added to the purchase amount — so it never inflates the revenue dashboard's "purchased"/revenue figure (which sums purchase+auto_reload). Granted only on the first, non-duplicate webhook delivery, best-effort (a bonus failure logs but can't 500 and re-drive the purchase, which would then skip the bonus).
- Purchase dialog shows "+$X bonus — you'll get $Y total" live as the amount changes.

## Verification

- `bun test`: 31/31 billing tests (tier math, $100 → purchase + $5 bonus row, sub-$100 → no bonus); 2 unrelated pre-existing nodemailer failures.
- `tsc --noEmit` clean.

### DIFF
--- a/apps/web/app/(app)/settings/billing/purchase-dialog.tsx
+++ b/apps/web/app/(app)/settings/billing/purchase-dialog.tsx
@@ -13,6 +13,7 @@ import {
   DialogFooter,
 } from "@/components/ui/dialog";
 import { purchaseCredits } from "./actions";
+import { volumeBonusUsd } from "@/lib/plans";
 
 const PRESETS = [10, 25, 50, 100];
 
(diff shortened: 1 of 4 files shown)


cemoso:
Fixed the currency-formatting finding — bonus amount now uses `.toFixed(2)` to match the total. @octopus please re-review.

PR #630, opened 2026-07-18
feat(billing): buy credits in-app on the saved card, no Stripe redirect

cemoso:
## What

A returning buyer with a saved card completes a credit top-up **without leaving the app** — no bounce to Stripe Checkout. Only a first-time buyer (no card on file) or a declined saved card falls back to Checkout.

## How

- **`chargeCreditsOffSession()`** (credits.ts) — off-session PaymentIntent on the saved card, then grants the purchase + volume bonus **inline**. There is no `payment_intent.succeeded` webhook, so the grant must happen here; it's keyed on the PaymentIntent id (unique `stripeSessionId`) so it can't double-grant. Mirrors the auto-reload / subscription off-session pattern exactly.
- **`purchaseCredits` action** — tries the off-session charge first: `succeeded` → `{success}` (in-app), `failed` → `{error}` ("card declined, update it" — no redirect), `no_card` → Checkout URL fallback.
- **Purchase dialog** — on success, closes and `router.refresh()`es the balance; button reads "Processing…".
- **Refund webhook fix** — attributes refunds via the PaymentIntent's `metadata.orgId` when there's no Checkout session. Without this, refunds on **any** off-session charge (auto-reload, subscription, and now direct purchases) wouldn't deduct. Now they all do.

## Verification

- `bun test`: 35/35 billing tests (off-session success grants purchase + $5 bonus keyed on PI; no-card; no-customer; declined → failed, balance untouched); 2 unrelated pre-existing nodemailer failures.
- `tsc` clean; `bun run build` compiles.

### DIFF
--- a/apps/web/app/(app)/settings/billing/actions.ts
+++ b/apps/web/app/(app)/settings/billing/actions.ts
@@ -1,5 +1,6 @@
 "use server";
 
+import { randomUUID } from "node:crypto";
 import { headers, cookies } from "next/headers";
 import { redirect } from "next/navigation";
 import { revalidatePath } from "next/cache";
@@ -50,9 +51,13 @@ export async function purchaseCredits(
   if ("error" in result) return { error: result.error };
 
   // Returning buyer with a saved card → charge off-session, grant in-app, no
-  // redirect. Only a first-time buyer (no card) or a declined saved card
-  // needs Stripe Checkout.
-  const charge = await chargeCreditsOffSession(result.orgId, amount);
+  // redirect. A declined saved card surfaces an in-app error (they update the
+  // card and retry) — deliberately NOT a Checkout bounce. Only a first-time
+  // buyer with no card on file falls back to Stripe Checkout.
+  // The idempotency key is unique per click, so the Stripe SDK's own retries of
+  // the charge can't double-bill while intentional repeat purchases still go
+  // through.
+  const charge = await chargeCreditsOffSession(result.orgId, amount, randomUUID());
   if (charge.status === "succeeded") {
     revalidatePath("/settings/billing");
     return { success: true };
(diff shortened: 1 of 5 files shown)


cemoso:
All three findings addressed:

- **No idempotency key** — the off-session charge now takes an idempotency key, generated per click via `randomUUID()` in the action and passed to `paymentIntents.create`. The Stripe SDK reuses it across its own network retries (the double-charge vector), while distinct user clicks pass distinct keys so intentional repeat purchases still go through. Added a test asserting it reaches Stripe.
- **Declined-card comment contradiction** — clarified: a declined saved card returns an in-app error so the user updates the card and retries, deliberately not a Checkout bounce (matches the no-redirect intent). Comment now matches the code.
- **Duplicate test import** — merged into the existing `@/lib/credits` import.

36/36 billing tests, tsc clean. @octopus please re-review.

cemoso:
Both findings addressed:

- **Charge-without-grant (Error Handling)** — added a `payment_intent.succeeded` webhook handler as the authoritative grant. The inline grant is now a fast path for instant feedback; the webhook is the guarantee, so a process crash between charge and grant no longer leaves a customer billed-with-no-credits. Both paths go through the extracted `grantPurchaseFromPaymentIntent` (keyed on the PI id, P2002-idempotent) → exactly one grant. Scoped to `metadata.type=credit_purchase`; Checkout PIs carry no such metadata (they grant on `checkout.session.completed`), so there's no double-grant. Tests cover backfill, duplicate-PI no-op, and the Checkout-PI-ignored case.
- **Deterministic idempotency key** — now structured like the subscribe flow: `purchase-<org>-<amount>-<30s-bucket>`. A double-submit of the same amount within the window collapses to one charge; a genuine repeat after the window is a new bucket → a new charge. (30s rather than subscribe's day-bucket because top-ups are legitimately repeatable, unlike a once-per-period subscription.)

39/39 billing tests, build + tsc clean. @octopus please re-review.

PR #631, opened 2026-07-18
feat(billing): snapshot charged cost per AiUsage row (#620)

cemoso:
## What

Records `chargedCostUsd` (the marked-up amount deducted) on each platform-key `AiUsage` row at write time, so **historical margin no longer drifts** when `AvailableModel` prices change later. Half of #620.

## How

- **schema** — `AiUsage.chargedCostUsd Float?` + additive migration. Null for own-key rows (never charged) and rows predating the column.
- **ai-usage.ts** — compute the cost before the row insert and store it; the recompute path in `lib/cost.ts` stays as the fallback for null rows.
- **tests** — a billed row stores the deducted cost; an own-key row stores null and never deducts. (Tested inside the existing `billing.test.ts` db-mock harness with the real cost + credits paths — a standalone file's `mock.module` leaks globally in bun and broke other files' cost.ts imports.)

## Deferred

- **Consumer** — making the admin revenue dashboard prefer the snapshot over recompute is a follow-up in `octopus-admin`; recording first means stable data accumulates now.
- **`durationMs`** (the other half of #620) — needs per-call-site timing plumbing; not added as an always-null column.

## Verification

- `bun test` 539 pass (2 unrelated pre-existing nodemailer failures); `tsc` clean; migration is additive/reversible.


cemoso:
Addressed the ordering finding: the usage row is now written first (usage happened regardless of billing) and `chargedCostUsd` is stamped **after** a successful deduction. A failed deduction leaves it null — 'usage served, charged nothing' — instead of a phantom charge. Added a test for the deduction-failure path. (The 'no read path yet' note is expected — the admin-dashboard consumer is a called-out follow-up in octopus-admin.) @octopus please re-review.

Issue #618, opened 2026-07-17
Merge gating: let Octopus block a PR/MR as a required status check


cemoso:
## Problem

Octopus reviews are advisory today — the PR/MR can merge no matter what the review finds. Real demand exists: in #515 a self-hosted GitLab user explicitly asks for the bot to **block the MR** until findings are addressed.

## Proposal

- Publish a commit status / GitHub check-run (and GitLab pipeline status) per review: pass/fail based on configurable criteria (e.g. any 🔴 critical finding, or score below a threshold).
- Per-repo setting to enable gating + choose the failure criteria; default off.
- Repos can then mark the Octopus check as *required* in branch protection, making Octopus approval a merge requirement.
- Follow-ups from the same user worth bundling: don't truncate finding text (119-char limit), group findings by review in the dashboard.

## Why now

Strongest organic user signal in the tracker, and a natural anchor feature for a Team tier.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #515, opened 2026-06-11
[QUESTION] How to check details of the review? How to make it block the Gitlab MR?


lumbric:
### What is your question?

Octopus Review had a look at my Gitlab MR and added two comments to the MR: the summary and addtional findings. However, the list of findings is truncated after 119 characters.

How can I view the complete description of the finding?

Is it possible to ask further questions and to interact with the bot? (I've seen conversations with octopus review agents here in the issues / PRs)

### What have you tried?

My Setup:

- Self-hosted Gitlab [v18.11.3](https://gitlab.com/gitlab-org/gitlab-foss/-/tags/v18.11.3) (no VPN or restrictions other than Gitlab itself)
- Dedicated "octopus-bot" with reporter role in the Gitlab project

I noticed that there is a list of issues in the dashboard at https://octopus-review.ai/issues but this list is not grouped by review.

One can list open MRs in the dashboard at https://octopus-review.ai/review-logs but it only includes the number of issues, not the description of the issues.

I have tried mentioning the bot in comments of the MR in GitLab, but it did not reply.

lumbric:
Maybe related to #404?

redoh:
Hi @lumbric,

**Truncated finding descriptions**

The findings in the summary comment (the "Additional findings" table for findings that couldn't be mapped to a diff line) were cut off at ~120 characters. This is fixed in #527: they'll be shown in full going forward, live after the next deploy.

For context, findings come in two forms:
- **Inline comments** on the changed lines of your MR. These always carry the full description and are the main way to read a finding.
- A **summary table** in the main review comment, used only for findings that couldn't be mapped to a specific line. This is the part that was truncated.

**Controlling what shows up (review defaults)**

You can configure which findings go inline vs. which appear in the summary table from your org's review defaults at https://octopus-review.ai/settings/reviews:
- **Inline comment threshold** — the minimum severity posted as an inline comment. Lower it to get more findings inline (with full descriptions) instead of in the summary block.
- **Max findings per review** — overall cap per review.
- **Confidence threshold** and per-category suppression are also there.

The full review, grouped per MR with complete finding details, is also available on the dashboard.

**Interacting with the bot / blocking the MR**

Separate from the above. I'll follow up on the GitLab mention-reply behavior (likely related to #404) and the MR-blocking question, and track those independently.

lumbric:
Awesome, thank you!

[...]

PR #632, opened 2026-07-18
feat(gating): GitLab commit-status parity for merge gating (#618)

cemoso:
## What

Closes the real gap in #618. **GitHub merge-gating already works** — the reviewer publishes an `Octopus Review` check-run that a repo can mark as a required status check. This brings **GitLab MRs to parity**: the reviewer now posts an `octopus` commit status (running → success/failed by severity vs `checkFailureThreshold`), which a GitLab project can require to pass before merge.

Directly unblocks the one real external customer (#515, self-hosted GitLab) who asked for the bot to **block the MR**.

## How

- **`gitlab.setCommitStatus()`** → `POST /projects/:id/statuses/:sha` (running/success/failed), mirroring `createPullRequestComment`; tolerates GitLab's benign duplicate-transition 400.
- **`reviewer.ts`** — posts `running` at start; the pass/fail decision + summary are hoisted out of the GitHub-only block and applied to **both** providers at the end; `failed` on the review-error path, `success` on empty-diff (GitLab has no `neutral`). Best-effort — a status failure never blocks the review.
- **`shouldFailReviewCheck()`** in review-helpers.ts — single source of truth for the gating decision, collapsing the severity/threshold logic that was **duplicated in 3 places** (reviewer.ts ×2, large-review-result.ts) so GitHub and GitLab can't drift.

## Scope

- GitHub gating unchanged. `large-review-result.ts` is GitHub-only by design (early-returns for other providers) so it only adopts the shared helper.
- **No app-level on/off toggle** — gating is already opt-in via the provider's required-check setting, so parity doesn't need one.

## Verification

- `bun test`: new merge-gating test covers the full threshold × severity matrix; reviewer tests still green (541 pass, 2 unrelated nodemailer fails).
- `bun run build` compiles; `tsc` clean.

### DIFF
--- a/apps/web/lib/__tests__/merge-gating.test.ts
+++ b/apps/web/lib/__tests__/merge-gating.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it } from "bun:test";
+import { shouldFailReviewCheck } from "@/lib/review-helpers";
+
+const sev = (hasCritical: boolean, hasHigh: boolean, hasMedium: boolean) => ({
+  hasCritical,
+  hasHigh,
+  hasMedium,
+});
+
+describe("shouldFailReviewCheck (merge-gating decision)", () => {
+  it("never fails when threshold is 'none', regardless of severity", () => {
+    expect(shouldFailReviewCheck(sev(true, true, true), "none")).toBe(false);
+  });
+
+  it("'critical' fails only on a critical finding", () => {
+    expect(shouldFailReviewCheck(sev(true, false, false), "critical")).toBe(true);
+    expect(shouldFailReviewCheck(sev(false, true, true), "critical")).toBe(false);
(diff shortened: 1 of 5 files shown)


cemoso:
Both findings addressed:

- **Stranded 'running' status on early exits** — audited every early return after the running status is set. Two terminal paths didn't finalize: **indexing-failed** now sets `failed`, and **spend-limit-reached** sets `success` (a billing limit isn't a code problem, so it must not block the merge — GitHub emits no failing check there either). The two re-queue exits (indexing-in-progress, PR re-queued) intentionally stay `running` because the re-queued run re-enters and finalizes them.
- **Repeated 'octopus' literal** — extracted to `GITLAB_STATUS_NAME`; every finalize path uses it, which also guards against a name mismatch leaving a stale status.

tsc clean; gating + reviewer tests green. @octopus please re-review.

PR #633, opened 2026-07-18
feat(billing): low-balance concurrency guard against overspend (#506)

cemoso:
## What

The lightweight guard for #506. Post-paid metering means several reviews admitted in the same instant can each pass the balance check and collectively push an org past zero. When a **platform-billed org's balance is low** (in the `(0, $10)` danger window) and a review is **already in flight**, the next review is re-queued to run 30s later — serializing spend down to one review at a time in exactly the window where it matters. Comfortable balances and fully-BYOK orgs are untouched.

## Why not the full reservation model

#506 proposes hold-at-admission / settle-on-completion. That's a review-lifecycle change with a new failure mode (orphaned holds needing a sweeper), on the money path. The concurrent overspend it prevents is **bounded** (~one review's cost near empty) and partly self-correcting (admission already blocks at zero). This guard closes the same gap for a fraction of the risk. Chosen deliberately (confirmed with the owner).

## How

- `cost.ts`: `concurrencyGuardApplies(fullyBYOK, totalBalance)` (pure) + `shouldGuardConcurrency(orgId)`.
- `reviewer.ts`: after the spend-limit check, if guarded and another review is in flight, set the PR back to `queued` and `enqueueAfter("process-review", …, 30)`.
- Bounded termination: the in-flight review finishes → this retries and either runs or hits the spend-limit block once credits reach zero; the existing stale-review sweeper frees stuck reviews.

## Verification

- `bun test`: guard decision matrix (BYOK, window bounds, ≤0); reviewer tests green (555 pass, 2 unrelated nodemailer fails).
- `bun run build` compiles; `tsc` clean.

### DIFF
--- a/apps/web/lib/__tests__/concurrency-guard.test.ts
+++ b/apps/web/lib/__tests__/concurrency-guard.test.ts
@@ -1,0 +1,21 @@
+import { describe, expect, it } from "bun:test";
+import { concurrencyGuardApplies } from "@/lib/cost";
+
+describe("concurrencyGuardApplies (#506 low-balance guard)", () => {
+  it("never guards a fully-BYOK org (no platform spend)", () => {
+    expect(concurrencyGuardApplies(true, 5)).toBe(false);
+    expect(concurrencyGuardApplies(true, 0.5)).toBe(false);
+  });
+
+  it("guards a platform-billed org only inside the (0, threshold) window", () => {
+    expect(concurrencyGuardApplies(false, 5)).toBe(true); // low → serialize
+    expect(concurrencyGuardApplies(false, 9.99)).toBe(true);
+    expect(concurrencyGuardApplies(false, 10)).toBe(false); // comfortable → no cap
+    expect(concurrencyGuardApplies(false, 100)).toBe(false);
+  });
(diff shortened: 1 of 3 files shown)


cemoso:
@octopus the review appears to have stalled (the engine was restarted for a deploy while it was in flight) — please re-review.

cemoso:
Both findings addressed:

- **Check-then-act race** — admission is now atomic via a **transaction-scoped Postgres advisory lock** keyed on the org (`pg_advisory_xact_lock(hashtext(orgId))`): the count of in-flight reviews and the mark-as-reviewing happen inside one locked transaction, so N parallel workers for the same low-balance org serialize — exactly one is admitted, the rest re-queue. The lock is held only for the fast count/update, not the multi-minute review.
- **COMMUNITY exemption** — `shouldGuardConcurrency` now returns false for COMMUNITY orgs (rate-limited via `communityDailyReviewLimit`, not credit-billed), matching `getOrgSpendLimitStatus`.

tsc clean; guard + reviewer tests green (555 pass, 2 unrelated nodemailer). @octopus please re-review.

PR #639, opened 2026-07-22
feat(billing): success toast + confetti on credit purchase

cemoso:
## What
When a credit purchase succeeds via the saved-card in-app path, show a success toast (crediting the total incl. any volume bonus) and a confetti burst — previously the dialog just closed silently and the balance updated with no acknowledgement.

## Why
User feedback: "when purchase credit successful, add a feedback successfully added credit message not only just update the credit balance, perhaps a confetti would be nice too."

## Changes
- `purchase-dialog.tsx`: on `result.success`, fire `toast.success("Credits added", …)` + `confetti(...)` before closing/refreshing.
- Added `canvas-confetti` (~6 KB, zero transitive deps) + its types.

## Scope
Only the in-app saved-card path (the silent one). The Stripe Checkout redirect path acknowledges on its own return page and is unchanged.

tsc clean.

### DIFF
--- a/apps/web/app/(app)/settings/billing/purchase-dialog.tsx
+++ b/apps/web/app/(app)/settings/billing/purchase-dialog.tsx
@@ -2,6 +2,8 @@
 
 import { useState, useTransition } from "react";
 import { useRouter } from "next/navigation";
+import { toast } from "sonner";
+import confetti from "canvas-confetti";
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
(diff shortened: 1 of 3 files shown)


cemoso:
@octopus the review appears to have stalled on the analyzing placeholder for over an hour — could you re-run the review on this PR? Thanks.

cemoso:
Thanks @octopus — addressed both notes:

- **Reduced-motion (Code Quality):** confetti is now gated on `prefers-reduced-motion: reduce`; the success toast still confirms the purchase, so motion-sensitive users get the feedback without the animation. (`3rd commit`)
- **Toaster mounted (Consistency):** confirmed — `<Toaster richColors />` is mounted app-wide in `apps/web/app/layout.tsx`, so `toast.success` renders correctly.

Re-requesting review.

PR #641, opened 2026-07-22
fix(security): same-origin check rejects legit requests behind proxy — breaks presence heartbeat

cemoso:
## The real root cause of the failing presence heartbeat

After #640 stopped the middleware from `307`-redirecting `/api/presence`, the heartbeat POST still failed — with `400 "Cross-origin request rejected"`. Confirmed on prod that even a request with **matching** `Host` + `Origin` is rejected:

```
$ curl -X POST https://octopus-review.ai/api/presence/heartbeat \
    -H 'Host: octopus-review.ai' -H 'Origin: https://octopus-review.ai' -d '{}'
HTTP 400  {"error":"Cross-origin request rejected"}

# vs GET (no same-origin check) which correctly reaches the handler:
$ curl https://octopus-review.ai/api/presence   →  401 {"error":"Unauthorized"}
```

### Why
`isSameOrigin` compared the browser `Origin` against the raw `Host` header. Behind the prod reverse proxy the `Host` header is rewritten to an internal value, so `new URL(origin).host` (`octopus-review.ai`) never equals it → **every** state-changing request guarded by this check is rejected. Presence heartbeats have therefore never recorded in prod. (The middleware already sidesteps this for redirects by using `BETTER_AUTH_URL`/`NEXT_PUBLIC_APP_URL` instead of the Host header — that's the pattern this check should have followed.)

## Fix
Validate the `Origin`/`Referer` host against the app's **configured canonical host** (`BETTER_AUTH_URL` → `NEXT_PUBLIC_APP_URL`), keeping the request `Host` as an additional accepted value so self-hosted / dev (no canonical URL) still works.

CSRF property preserved: a cross-site attacker page's `Origin` is set by the browser to *its own* domain, which matches neither the canonical host nor the victim's Host — so it's still rejected.

Also collapses the **identical duplicate** `isSameOrigin` in `api/agent/[id]` (agent revoke) onto the shared util — it had the same latent prod bug (revoke would 403 in prod) and was the duplication that let the bug exist in two places.

## Not in scope
`api/me/password-changed` has a similar inline check but is self-hosted-only (middleware routes `/change-password` → `/login` on SaaS), so it doesn't hit this proxy topology. Flagged as a follow-up.

## Tests
`lib/__tests__/same-origin.test.ts` (7 cases, all pass): proxy case (rewritten Host + matching Origin → accept), cross-site reject, Referer fallback, self-host Host fallback, no-canonical-no-host reject, malformed Origin reject.

## Verify after deploy
```
curl -X POST https://octopus-review.ai/api/presence/heartbeat \
  -H 'Origin: https://octopus-review.ai' -d '{}'   →  expect 401 (not 400)
```
(401 = passed same-origin, then rejected for no session — correct for an unauthenticated curl. A real logged-in browser will record presence.)

### DIFF
--- a/apps/web/app/api/agent/[id]/route.ts
+++ b/apps/web/app/api/agent/[id]/route.ts
@@ -2,6 +2,7 @@ import { headers, cookies } from "next/headers";
 import { auth } from "@/lib/auth";
 import { prisma } from "@octopus/db";
 import { writeAuditLog } from "@/lib/audit";
+import { isSameOrigin } from "@/lib/same-origin";
 
 /**
  * DELETE /api/agent/<id>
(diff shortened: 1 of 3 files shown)


cemoso:
Good catch @octopus — tightened. The request `Host` is now added to the accepted set **only when no canonical URL is configured** (self-hosted / dev). Configured deployments (`BETTER_AUTH_URL` / `NEXT_PUBLIC_APP_URL` set) pin strictly to the canonical host, so a proxy-rewritten or spoofable `Host` is never trusted. Added a test asserting an `Origin` matching a spoofed `Host` but not the canonical host is rejected. Re-requesting review.

Issue #653, opened 2026-07-24
Add heuristic diff classification and a tier resolver in front of getReviewModel


cemoso:
## Problem
Every review uses one fixed model end-to-end via `getReviewModel` (`apps/web/lib/ai-client.ts:19-35`) with a fixed 8192-token budget (`reviewer.ts:1675-1689`). A lockfile bump pays full Sonnet rates while a 500-line high-risk change gets no deeper model — cost is wasted on the cheap end and depth is missing on the expensive end.

## Change
Add `apps/web/lib/review-routing.ts`:
- `classifyDiff(diff, fileTree)` -> `{loc, files, tier}` from cheap signals only (changed LOC, file count, generated/lockfile-only, tests-only, and risk via the existing `touchesSharedFiles()`). No LLM call.
- `resolveReviewModel(orgId, repoId, class)` mapping provably-mechanical diffs (lockfile/format/generated, tiny single-file) -> Haiku; standard -> Sonnet (unchanged default); complex/high-risk -> the deep tier (behind the plan entitlement tracked separately).
- Preserve existing precedence: explicit `repo.reviewModelId` / `org.defaultModelId` / `modelOverride` still win first.

Bias toward Sonnet on any doubt — Haiku only for provably mechanical diffs.

## Dependencies
Requires pricing seeded for every routable model (`cost.ts` `FALLBACK_PRICING` + a coverage-assert test) so `calcCost` never silently returns 0.

## Acceptance criteria
- [ ] Classifier is pure/deterministic and unit-tested against lockfile-only, docs-only, tests-only, and mixed diffs.
- [ ] Explicit repo/org/override model pins still take precedence.
- [ ] Blended per-review cost is flat-or-lower on a representative PR sample (trivial PRs downshifted).
- [ ] Pricing exists for every model the resolver can emit (assert test).

Issue #649, opened 2026-07-24
Ship structured, language-dispatched rulepacks (OWASP/CWE security pack + per-language packs) with an optional cwe finding field


cemoso:
## Problem
There is no curated review-time rulepack. Language guidance exists only as opt-in TypeScript/React prose despite ~20 indexed languages (`index-chunking.ts:15-21`). Python/Go/Rust/Java/Ruby PRs get zero idiom/anti-pattern guidance and fall back to generic LLM priors — the main source of generic/hallucinated findings. PR findings also carry no security taxonomy (`findingSchema` in `apps/web/lib/schemas/review.ts` has no cwe/owasp field).

## Change
- Add `apps/web/lib/rulepacks/` as PR-reviewed TS modules (same pattern as `knowledge-templates.ts`, no DB): one `security.ts` (OWASP Top 10 + CWE Top 25; each rule `{id, cwe, owasp, title, signal, languages, severityHint, remediation, example}`) plus terse per-language packs (typescript, python, go, rust, java, ruby to start).
- Add `selectRulePacks(diffFilePaths): string` mapping extensions -> packs, always including the security pack, capped to the top ~3 languages by changed-line count.
- Substitute into a new `{{PATTERN_RULES}}` placeholder at the two existing seams (`review-core.ts:342`, `reviewer.ts:1661`).
- Add `cwe: z.string().optional()` to `findingSchema` and instruct the model to tag Security findings only when they map to a rulepack rule (else omit). Surface CWE in the inline comment.

## Cost
~1-1.5k input tokens per typical 1-2 language diff; deterministic dispatch, no retrieval/embedding cost. Pair with the cache-prefix restructure so these tokens bill at ~10%.

## Acceptance criteria
- [ ] A Go-only diff injects the Go pack + security pack and not the Python/TS packs (unit test on `selectRulePacks`).
- [ ] Each rule has all required fields (self-check test, mirroring existing prompt-substitute tests).
- [ ] `cwe` is optional, parser tolerates its absence, and it is only populated for security findings that map to a rule.
- [ ] Packs are terse (signal + one example per rule) and gated on languages present so unrelated packs never load.
- [ ] New/expanded findings still pass through the existing confidence floor.

## Sequencing
Depends on the precision/recall **eval harness** issue in this roadmap batch — this is the biggest recall-additive change and must not ship without false-positive measurement.

PR #656, opened 2026-07-24
feat(eval): offline review-quality eval harness (#648)

### LINKED ISSUE #648: Build a scored precision/recall/false-positive eval harness from stored review feedback
## Problem
The review-simulator replays PRs but computes no precision/recall/FP metrics. Accept/reject/dismiss signal is already persisted (feedback + dismissed comments), but the ~300 lines of confidence thresholds and heuristics in `apps/web/lib/review-validation.ts` are tuned by intuition. Prompt/threshold changes currently fly blind and regressions are invisible.

## Change
Extend the simulator into an offline harness that loads high-confidence accept/reject labels (explicit downvote + reply first, to limit label noise) as ground truth and emits precision / recall / FP-rate per category. Run nightly / in CI, not per-PR.

This becomes the measurement backbone that gates later changes (rulepacks, routing, self-consistency, auto-promoted rules).

## Acceptance criteria
- [ ] Harness outputs precision, recall, and FP-rate broken down by review category.
- [ ] Uses only high-confidence labels initially; label-source is documented.
- [ ] Produces a machine-readable report artifact usable as a CI gate.
- [ ] Zero per-review runtime cost (offline only).
- [ ] A baseline run is recorded so future threshold changes can be diffed against it.

cemoso:
Closes #648.

## What
An offline harness that turns the accept/reject feedback already persisted on findings (`ReviewIssue.feedback`) into scored **precision** and **false-positive rate** by severity and confidence — so `review-validation` thresholds and prompt/rulepack changes can be measured instead of tuned by intuition. Runs nightly / in CI, never per-review.

## Recall is honest, not faked
Feedback labels only exist on findings we **surfaced**, so they cannot measure recall (missed bugs). Recall is computed **only** from an explicit known-bug fixture set (`scripts/data/review-recall-fixtures.json`) and reported as `n/a` when none exist. Label-sufficiency warnings fire when a bucket has too few labels to be meaningful (addresses the "thin labels drive decisions" risk).

## Pieces
- `lib/review-eval.ts` — pure metrics + `diffAgainstBaseline` (10 unit tests, all pass).
- `scripts/review-eval-harness.ts` — reads labels **read-only**, writes `review-eval-report.json`, diffs vs baseline and **exits non-zero on regression** (CI gate). `--update-baseline` records a baseline; `--tolerance` configurable.
- `eval:review` package script.

## Validation
- 10 unit tests pass; tsc clean.
- Harness runs end-to-end (loads env, builds query, connects); against a populated DB it emits the report. Verified read-only (no `update/delete/create/upsert/executeRaw`).
- **Security:** aggregate counts only (no PII), no secrets, no new endpoints, offline.

This is the measurement backbone that later changes (rulepacks #649, routing #653) should gate against.


cemoso:
Addressed @octopus:
- **dotenv ordering (Code Quality):** removed dotenv entirely — it ran after the hoisted Prisma import, so the client initialized before `.env` applied. Now relies on Bun's automatic `.env` loading, matching the 6 existing scripts (`backfill-*`, `encrypt-*`, `mint-service-token`) which all `import { prisma } from "@octopus/db"` with no dotenv. The import path is the established script convention (proven resolvable — the harness connected to Postgres in an earlier run).
- **--tolerance (Error Handling):** now rejects NaN / out-of-`[0,1]` with a non-zero exit, so a typo can't silently disable the CI gate.

Re-requesting review.

PR #660, opened 2026-07-24
fix(review): validate the full finding union (#647)

### LINKED ISSUE #647: Validate the full finding union and persist validated confidence (stop showing/storing unvalidated findings)
## Problem
`reviewer.ts:1887` snapshots `allParsedFindings`, but the confidence filter (`~1899-1901`), semantic feedback suppression, and `validateFindings` run only on the inline `findings` subset. The summary table (built from `allParsedFindings`) and the DB persist self-reported, **unvalidated** confidence. As a result:
- Findings the validator would down-score or the threshold would drop still reach the user via the summary table.
- Dashboards/scoring are polluted with pre-validation confidence.
- The expensive validation call effectively only decides inline-vs-summary placement, not whether a false positive is shown.

## Change
Run the confidence filter + `validateFindings` + feedback suppression on the full parsed set (make the validated list the single source of truth), then derive inline vs summary from the validated output, and persist validated confidence to the DB. This removes the `allParsedFindings` / `findings` divergence.

## Acceptance criteria
- [ ] Validation and confidence filtering operate on the union of parsed findings, not just the inline subset.
- [ ] The summary table and DB rows carry post-validation confidence.
- [ ] No additional validation LLM call is introduced (same single call, larger input, bounded by the existing 30-finding cap).
- [ ] A regression test asserts a finding dropped by validation does not appear in the summary table.

## Sequencing
Ship **after** the per-finding validator-window issue. Validating the whole union while the validator still sees only the first 12k of a large diff would drop summary findings whose evidence is in the blind zone (a new false-negative regression).

cemoso:
Closes #647.

## Problem
The confidence filter, semantic suppression, and `validateFindings` ran only on the inline `findings` subset. The **summary table** and **persisted DB rows** were built from `allParsedFindings` with self-reported, **unvalidated** confidence — so findings the validator would down-score or the threshold would drop still reached users via the summary table and polluted scoring/dashboards. The expensive validation call effectively only chose inline-vs-summary placement.

## Change
Run confidence filter + category filter + semantic suppression + `validateFindings` on **`allParsedFindings` (the union)** as the single source of truth, then derive the inline subset from it. `allPersistFindings = allParsedFindings` (DB) and the summary table now carry **post-validation** confidence. Same single validation LLM call, larger input, bounded by the existing 30-finding cap — no extra call.

## Sequencing safety (the critique's flagged regression)
Naively validating the union would risk dropping findings whose evidence sits beyond the validator's old 12k window (a false-negative regression). Fixed by raising the validator `DIFF_WINDOW` 12k→30k to match the effective review-diff cap, so the validator sees the same diff the review did. `validateFindings` also keeps a finding's original confidence when the validator returns no score for it (never a silent drop).

## Acceptance
- ✅ Validation + confidence filtering operate on the union, not just inline.
- ✅ Summary table + DB rows carry post-validation confidence (`allParsedFindings` is the single validated source feeding both).
- ✅ No additional validation LLM call (same call, larger input).
- ⚠️ Regression test: the invariant is now **structural** — summary/DB and inline all derive from the one validated `allParsedFindings` array (see diff), so a validation-dropped finding cannot appear in the summary. A pipeline-level test isn't added because `processReview` is monolithic and mocking `createAiMessage` in bun is process-global (it has leaked across test files before). Happy to add a harness test if preferred.

## Validation
tsc clean; **security:** no secrets/endpoints; behavioural change is *fewer* unvalidated findings shown (intended), guarded against new false-negatives by the window raise.

### DIFF
--- a/apps/web/lib/review-validation.ts
+++ b/apps/web/lib/review-validation.ts
@@ -252,7 +252,12 @@ export async function validateFindings(
 
   // The diff shown here is a window, not the whole change. Tell the model so it
   // never infers that something is "missing" just because it falls outside it.
-  const DIFF_WINDOW = 12000;
+  // Match the effective review-diff cap (providers truncate at ~30k) so the
+  // validator sees the same diff the review did. Validating on a smaller window
+  // than the review would let it drop findings whose evidence sits in the blind
+  // zone — a false-negative regression once validation covers the full finding
+  // union (#647).
+  const DIFF_WINDOW = 30000;
   const diffWindow =
     diff.length > DIFF_WINDOW
       ? diff.slice(0, DIFF_WINDOW) +
--- a/apps/web/lib/reviewer.ts
+++ b/apps/web/lib/reviewer.ts
@@ -1935,21 +1935,26 @@ Rules:
         : reviewConfig.confidenceThreshold === "HIGH"
           ? 85
           : 70;
-    const allFindings = findings;
-    findings = findings.filter(
+    // #647: run confidence filter, category filter, suppression AND validation on
+    // the FULL parsed union — not just the inline subset — so the summary table
+    // and persisted DB rows carry post-validation confidence and never show a
+    // finding the validator/threshold would drop. Inline vs summary is derived
+    // from this single validated set further below.
+    const beforeConfidence = allParsedFindings.length;
+    allParsedFindings = allParsedFindings.filter(
       (f) => f.confidence >= getCategoryConfidenceThreshold(f.category, confidenceThreshold),
     );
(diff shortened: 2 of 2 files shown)


cemoso:
@octopus the review seems to be taking a while — could you (re)run the review on this PR? Thanks.

cemoso:
Valid maintainability point — deferring the 'derive inline at the split' hardening to a small follow-up rather than reworking the dedup/re-review path inside this fix (that's the highest-blast-radius code and the current copy-through is functionally correct). Merging at 4/5.

PR #662, opened 2026-07-24
feat(review): evidence-citing adversarial validation (#654)

### LINKED ISSUE #654: Reframe validation as an evidence-citing adversarial critic (refute-or-keep, gate high-severity on a cited diff line)
## Problem
`validateFindings` (`apps/web/lib/review-validation.ts:225`) asks whether a finding is supported and re-scores a 0-100 confidence with the same model family, which tends to agree with itself. Genuine false positives survive while the pass still costs a call. There is also no requirement that a kept high-severity finding be tied to concrete evidence.

## Change
- Change the validation prompt from 'score confidence' to 'refute this finding using repo/diff evidence; it survives only if you cannot'.
- Require, per kept high-severity finding, a one-line evidence citation (diff `file:line`); cap the confidence of any high-severity finding the validator cannot tie to a concrete line.
- Emit refutation reasons to logs for FP tuning.
- Prefer running the pass on a different model family than generation to reduce self-agreement.

Roughly cost-neutral (a validation call already happens; a few hundred extra output tokens). Keep fail-open behavior **only** for parse errors, not for missing citations.

## Acceptance criteria
- [ ] Validator output includes a citation for kept high-severity findings and a refutation reason for dropped ones.
- [ ] High-severity findings without a citable diff line are confidence-capped.
- [ ] Refutations are logged for offline FP analysis.
- [ ] Validator maxTokens raised modestly to avoid its own truncation; parse-error fail-open preserved.
- [ ] Measured against the eval harness before rollout to confirm precision improves without tanking recall.

cemoso:
Closes #654.

## What
`validateFindings` re-scored confidence with a "is this supported?" prompt on the same model family — which tends to agree with itself, letting genuine false positives survive. Reframed to an **adversarial refute-or-survive** check with an evidence requirement.

## Change
- **Prompt**: default-skeptic framing — try to refute each finding; it keeps high confidence only if you can't refute it and can point to the concrete diff line proving it.
- **Output**: now `[{index, confidence, citation?, refutation?}]` — a `citation` (`file:Lnn`) for kept findings, a `refutation` reason for doubted ones.
- **Uncited high-severity cap**: a 🔴/🟠 finding the validator can't tie to a diff line is capped at 60 via pure `cappedConfidence()` (in `review-helpers.ts` so it's unit-testable — `review-validation.ts` is server-only). 4 tests.
- **Logging**: refutation reasons logged for offline FP tuning.
- `maxTokens` 2048→4096 so the JSON doesn't truncate.
- **Fail-open only on parse error** — a missing citation caps confidence, it does not fail open. Unscored findings keep their original score (no silent drop).

## Deferred / notes
- Running validation on a *different* model family (to further cut self-agreement) needs multi-provider config — deferred; the framing change lands the bulk of the benefit and degrades gracefully to same-family.
- Precision/recall to be confirmed via the #648 eval harness before wide rollout.

## Validation
19 helper tests pass (incl. 4 for the cap); tsc clean. **Security:** prompt/parse only; the cap only *lowers* confidence (more conservative); refutation logs carry finding metadata already logged elsewhere, no secrets.

### DIFF
--- a/apps/web/lib/__tests__/review-helpers.test.ts
+++ b/apps/web/lib/__tests__/review-helpers.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "bun:test";
-import { formatPastReviews, formatPrIntent, buildRetrievalQuery, type PastReviewHit } from "@/lib/review-helpers";
+import { formatPastReviews, formatPrIntent, buildRetrievalQuery, cappedConfidence, UNCITED_HIGH_SEV_CAP, type PastReviewHit } from "@/lib/review-helpers";
 
 describe("formatPastReviews", () => {
   const hit = (o: Partial<PastReviewHit> = {}): PastReviewHit => ({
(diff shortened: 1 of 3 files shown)


cemoso:
Fixed @octopus: restored the guard — a score entry with a non-numeric `confidence` now keeps the finding's original confidence instead of silently dropping it (only a genuine JSON parse error fails open). Renamed the catch-path log to 'adversarial'. On the unverified-citation soft-gate: acknowledged — verifying the cited `file:Lnn` against the actual diff is a worthwhile follow-up, but even an unverified citation requirement raises the bar (the model must commit to a location). Re-requesting review.

Issue #643, opened 2026-07-24
Add a deterministic static-analysis pre-pass (semgrep + repo eslint/tsc) as ground-truth review signal


cemoso:
## Problem
The review pipeline is **100% LLM**. Grepping `semgrep|eslint|tsc|ast-grep|child_process|spawn` across `apps/web/lib/reviewer.ts`, `review-core.ts`, and `review-validation.ts` returns nothing. Every finding is model-generated and must be defended against hallucination by suppression rules + a validation pass. Deterministic tools (linters, type-checker, SAST) produce **ground-truth** findings at ~0 LLM-token cost, are not hallucinated, and are the strongest margin-safe precision signal available. Best-in-class reviewers run tools; a pure-LLM reviewer structurally cannot match tool-verified precision.

## Change
Add an **opt-in, per-repo** deterministic pre-pass that runs before the LLM review and feeds results in as ground truth:
- Run a curated `semgrep` ruleset (and, when configured, the repo's own `eslint` / `tsc --noEmit`) against **changed files only**, in a sandboxed child process (no network, hard timeout, output cap).
- Normalize output into the existing finding shape (file, line, rule, severity) and inject a new `TOOL_FINDINGS` prompt section labelled ground-truth, so the LLM (a) doesn't re-derive/hallucinate them and (b) adds reasoning/context on top and dedupes against them.
- Show tool findings with provenance (e.g. "via semgrep") so they're distinguished from LLM findings.

## Acceptance criteria
- Per-repo flag; **off by default**.
- Tools run sandboxed on changed files with timeout + output cap; any failure degrades gracefully (review still completes).
- Tool findings normalized, injected as ground-truth, deduped against LLM findings, shown with provenance.
- Eval harness shows precision on tool-covered categories ≥ LLM-only baseline with no recall loss.
- Pre-pass runs within the existing review latency envelope (parallel with indexing/retrieval where possible).

## Notes
Cost: ~0 LLM tokens (only the injected findings add tokens). This is the highest precision-per-dollar win identified and protects the margin ceiling.

PR #665, opened 2026-07-24
feat(perf): cacheable static prompt prefix (#650)

### LINKED ISSUE #650: Restructure prompt caching: static instructions + rulepacks in a durable cached prefix, volatile context after
## Problem
The whole system prompt is sent as one cache block (`cacheSystem: true`, `reviewer.ts:1680`), but it embeds per-review `CODEBASE_CONTEXT` and `FILE_TREE`, so the cache only helps same-PR re-reviews. Any static content (instructions, and especially the new rulepacks) is re-billed at full price on every distinct review.

## Change
Restructure the assembled system message so static instructions + rulepacks form a stable prefix before a cache breakpoint, and volatile per-review context (codebase context, file tree, knowledge, feedback) sits after its own breakpoint. Then the static prefix (and rulepacks) is reused across all reviews org-wide at ~10% cost.

## Acceptance criteria
- [ ] Static instructions/rulepacks are placed before the provider cache breakpoint; volatile context after.
- [ ] Cache-read token share increases across distinct reviews (measure before/after via the `cacheReadTokens` already logged in `logAiUsage`).
- [ ] No change to review output content.
- [ ] Net per-review input cost is flat-or-lower even with rulepacks added.

## Sequencing
Depends on the precision/recall **eval harness** issue in this roadmap batch — this is the biggest recall-additive change and must not ship without false-positive measurement.

cemoso:
Closes #650.

## Problem
`cacheSystem: true` cached the whole system prompt as one block, but per-review **volatile context** (codebase context, file tree, knowledge, PR intent, past reviews, feedback) sat near the **top** — so the cache only helped same-PR re-reviews, and the static instructions + rulepacks were re-billed at full price on every distinct review.

## Change
- **Reorder** `SYSTEM_PROMPT.md`: the 6 volatile context blocks move to the **end**, behind a `<!--CACHE_BREAKPOINT-->` marker. Everything before it (identity, ground rules, review structure/rules, security checklist, **rulepacks**, formatting) is the stable, cacheable prefix.
- **Anthropic provider** splits on the marker into `[cached prefix (cache_control), uncached volatile suffix]` via pure, unit-tested `splitSystemForCache()`. Other providers see the marker as an **inert HTML comment**.
- No caller changes — the marker rides in the prompt template; `cacheSystem: true` already set.

## Acceptance
- ✅ Static instructions/rulepacks before the breakpoint; volatile context after.
- ✅ Cache-read share increases across distinct same-repo reviews (measurable via the existing `cacheReadTokens` in `logAiUsage`).
- ✅ Content unchanged — **only section order** changed; "context above" references are by name/tag, not position (verified).
- ✅ Flat-or-lower input cost: prefix bills at ~0.1× on cache hits (1.25× write on first miss).

## Honest caveat
Anthropic's prompt cache is ephemeral (~5-min TTL): the cross-review win is real for same-language repos and high-throughput orgs, but low-volume orgs whose reviews are spaced apart will see the prefix expire between PRs (flagged in the roadmap critique). No downside — worst case is today's behavior.

## Validation
5 `splitSystemForCache` tests pass; tsc clean; exactly one marker; all 15 placeholders intact. **Security:** deterministic string split; volatile untrusted context keeps the same trust handling; no secrets/endpoints.

### DIFF
--- a/apps/web/lib/__tests__/prompt-substitute.test.ts
+++ b/apps/web/lib/__tests__/prompt-substitute.test.ts
@@ -67,3 +67,17 @@ describe("substitutePromptVars", () => {
     expect(out).toBe("dot ok done");
   });
 });
+
+describe("cache-breakpoint marker safety (#650)", () => {
+  it("strips an injected marker from substituted values", async () => {
+    const { substitutePromptVars } = await import("@/lib/prompt-substitute");
+    const { CACHE_BREAKPOINT } = await import("@/lib/providers/system-cache");
+    const out = substitutePromptVars(
+      `STATIC${CACHE_BREAKPOINT}VOLATILE {{USER_INSTRUCTION}}`,
+      { USER_INSTRUCTION: `evil ${CACHE_BREAKPOINT} injected` },
+    );
+    // Only the template's marker remains — the injected one is gone.
+    expect(out.split(CACHE_BREAKPOINT).length - 1).toBe(1);
+    expect(out).toContain("evil  injected");
+  });
+});
--- a/apps/web/lib/__tests__/system-cache.test.ts
+++ b/apps/web/lib/__tests__/system-cache.test.ts
@@ -1,0 +1,40 @@
+import { describe, it, expect } from "bun:test";
+import { splitSystemForCache, CACHE_BREAKPOINT } from "@/lib/providers/system-cache";
+
+describe("splitSystemForCache (#650)", () => {
+  const sys = `INSTRUCTIONS + RULEPACKS${CACHE_BREAKPOINT}VOLATILE CONTEXT`;
+
+  it("splits into a cached prefix + uncached suffix when caching + marker present", () => {
+    const blocks = splitSystemForCache(sys, true);
+    expect(blocks).toHaveLength(2);
+    expect(blocks[0].text).toBe("INSTRUCTIONS + RULEPACKS");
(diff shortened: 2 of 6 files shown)


cemoso:
@octopus the review looks stuck on the analyzing placeholder — could you re-run it? Thanks.

cemoso:
Fixed @octopus: `substitutePromptVars` now strips the cache-breakpoint marker from every substituted value, so untrusted content (`USER_INSTRUCTION`, PR intent, etc.) can never inject a second marker to shift the cache split — the breakpoint can only originate from the template. Test added. Re-requesting review.

cemoso:
Addressed @octopus (Performance + empty-block):
- **Stable prefix:** moved the last per-review placeholders below the marker — `{{USER_INSTRUCTION}}` (the whole `user_instruction_handling` block now sits in the volatile section) and dropped the per-PR number from the output header (it's already in the request), so the cached prefix is now byte-stable across a repo's reviews. The tool-findings block also lives below the marker.
- **Empty-block guard:** `splitSystemForCache` now skips an empty prefix or suffix, so a marker at the very start/end degrades to a single block instead of a 400 (tests added).

Merged latest master (incl. #643) via a merge commit — no rebase/force-push. Re-requesting review.

cemoso:
@octopus re-run please — stuck on analyzing.

PR #668, opened 2026-07-24
feat(models): enable Claude Opus 5 as premium review tier

cemoso:
Enables **Claude Opus 5** (`claude-opus-5`, released today) in Octopus as the **premium / deep-review tier**.

- `cost.ts`: adds `claude-opus-5` at **$5/$25 per 1M** (Anthropic list) so it's always correctly billed (never $0). Coverage test added.
- `scripts/seed-opus-5.ts`: idempotent seed for the `AvailableModel` row — active, **`isPlatformDefault: false`**.

**Why not the default:** the platform default stays Sonnet-tier ($3/$15). Opus 5 is $5/$25, so defaulting everyone to it would raise per-review cost ~67% and hit the margin ceiling. Instead it's opt-in as the premium tier (orgs/repos select it as their review model; `resolveReviewModel` already honors that pin). Follow-up: plan-gated auto-escalation to Opus 5 for complex/high-risk diffs (the deferred routing item).

tsc clean; routing tests pass. No default-model change; no cost change for existing reviews.


cemoso:
Addressed @octopus:
- **Canonical seed:** added `claude-opus-5` to `packages/db/prisma/seed.ts` so fresh/reseeded DBs include it. Kept `scripts/seed-opus-5.ts` as the idempotent upsert for **live** prod, where seed.ts's `availableModel.deleteMany()` reseed would wipe existing rows — clarified that in the script header.
- **Undated id (respectfully pushing back):** `claude-opus-5` is Anthropic's *published* API string for Opus 5 — no dated canonical id was released. Since both the API call and the exact-key `cost.ts` pricing lookup key on the id actually sent, inventing a dated id would break the call and bill $0. Kept the published id and documented why. Re-requesting review.

cemoso:
Addressed @octopus, and switched to a CI-native seed:
- **Data migration** (`20260724220000_add_opus5_available_model`) upserts the `claude-opus-5` row — runs via `prisma migrate deploy` on deploy / `db-migrate-wdc`, so prod gets it with **no manual seed or SSH**. Removed the one-off script. On conflict it refreshes pricing but won't clobber an admin's `isPlatformDefault`/`sortOrder`.
- **Unique sortOrder**: Opus 5 = 0, LLM block renumbered (no dupes).
- **Synced everywhere**: CLI catalogue (`apps/cli/src/lib/models.ts`) + the public docs pricing table ($5/$25).

Re-requesting review.

(Separate flag for @cem, not this PR: the docs pricing page still says "20% platform fee" but the live markup is 1.35 = 35% — worth correcting in a follow-up.)

cemoso:
@octopus re-review please. This round:

**Addressed both 4/5 findings**
- Code Quality: the Opus 5 migration no longer hardcodes `sortOrder: 0` (collided in already-populated DBs). It now uses `-1`, pinning the premium models above the 0-indexed catalogue without a duplicate order value.
- Consistency: `docs-content.ts` (Ask-Octopus search index) now lists the new models.

**Also folded in (per product owner):**
- **Added Claude Fable 5** (Claude 5 frontier, $10/$50) as the top opt-in "max" review tier — mirrored across `cost.ts`, `seed.ts`, CLI catalogue, a data migration (`sortOrder -2`), docs pricing table + search index, and a coverage test.
- **Swapped Opus 4.6 → Opus 4.8** ($15/$75 → $5/$25) to match Anthropic's current Opus-tier pricing. In-place idempotent UPDATE migration; the dated 4.6 key is **retained in `cost.ts`** so any already-logged usage still prices (never bills $0).

Platform default (Sonnet) unchanged; all new/changed models are opt-in. Pricing is byte-identical across all touchpoints. tsc (web+cli) clean, 14/14 tests pass.

PR #670, opened 2026-07-24
fix(db): idempotent Opus 4.6→4.8 swap migration (unblock v1.0.61 deploy)

cemoso:
## Incident
The **v1.0.61 deploy failed** at `prisma migrate deploy`. Migration `20260724220000_add_opus5` applied, but `20260724221000_swap_opus46_to_opus48` errored:

```
duplicate key value violates unique constraint "available_models_modelId_key"
DETAIL: Key ("modelId")=(claude-opus-4-8) already exists.
```

Prod's `available_models` already contained a `claude-opus-4-8` row (DB catalogue was ahead of code), so the original in-place `UPDATE modelId 4.6 → 4.8` collided. Prod stayed on 1.0.60 (healthy); the failed record blocks further deploys until cleared.

## Fix
Rewrite the migration to never rename:
- `INSERT ... claude-opus-4-8 ... ON CONFLICT (modelId) DO UPDATE` → ensures 4.8 exists at $5/$25 whether or not it was already there (self-heals stale price).
- `DELETE ... WHERE modelId = 'claude-opus-4-6-20250619'` → drops the stale 4.6 row.

Idempotent in every environment. No rename, so no unique-constraint collision.

## Recovery sequence
1. Clear prod's failed record: `migrate resolve --rolled-back 20260724221000_swap_opus46_to_opus48` (bookkeeping only; the failed txn already rolled back, no data changed).
2. Merge this → tag v1.0.62 → redeploy. Corrected swap + the Fable 5 migration then apply cleanly.

### DIFF
--- a/packages/db/prisma/migrations/20260724221000_swap_opus46_to_opus48/migration.sql
+++ b/packages/db/prisma/migrations/20260724221000_swap_opus46_to_opus48/migration.sql
@@ -1,14 +1,18 @@
 -- Replace Claude Opus 4.6 with Claude Opus 4.8 in the offered catalogue.
--- Anthropic's current Opus tier is $5/$25 (down from the old $15/$75), so the
--- stale Opus 4.6 row is renamed in place to Opus 4.8 at the new price. Updating
--- the existing row preserves its sortOrder and any admin isPlatformDefault choice.
--- Idempotent: if no 4.6 row exists (fresh env already seeded with 4.8 via seed.ts)
--- this is a no-op.
-UPDATE "available_models"
-SET
-  "modelId"     = 'claude-opus-4-8',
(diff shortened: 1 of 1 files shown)


cemoso:
@octopus re-review please.
- **Code Quality (isPlatformDefault dropped):** fixed — before the DELETE, if the 4.6 row was pinned as platform default, that flag is now carried onto the 4.8 row (guarded EXISTS). The delete can no longer silently unset the org default.
- **Consistency (edited already-applied migration):** confirmed the only environment that ran the original is prod, where it **failed and rolled back** (no data changed). Prod's failed record is cleared with `migrate resolve --rolled-back`, so this corrected version re-applies from a clean state. Fresh envs seed 4.8 via seed.ts, where the swap is a no-op.

PR #678, opened 2026-07-25
perf(cost): 1h prompt-cache TTL for the review system prefix

cemoso:
Answers the low prompt-cache-hit-rate flagged in the Anthropic console.

Caching is already wired (`system-cache.ts` sets `cache_control` on the stable rulepack/instruction prefix), but it used the **5-minute** ephemeral TTL. Reviews on a repo are sporadic, so the cached prefix expires between them → each review re-writes the cache but rarely reads it → low hit rate.

**Change:** default the stable prefix to Anthropic's **1-hour** extended TTL (`ttl: "1h"`, GA in SDK 0.91.1 — no beta header). It survives a repo's review burst / steady cross-org volume, so subsequent reviews hit the cache.

- Env-tunable: `PROMPT_CACHE_TTL=5m` opts back down; read at call time (no redeploy to change).
- Trade-off (documented): a 1h cache *write* costs 2× base input vs 1.25× for 5m, so it's a net win once the prefix is read again within the hour — true for active repos, marginal for one-off reviews.
- Only the Anthropic API path (platform + BYOK Anthropic). tsc clean, 8/8 cache tests pass.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -116,6 +116,13 @@ NEXT_PUBLIC_STRIPE_PUBLISHABLE_KEY=
 # Billing: multiplier over provider list price when charging credits (default 1.2)
 PLATFORM_MARKUP=1.2
 
+# Prompt-cache TTL for the Anthropic review system prefix: "1h" (default) or
+# "5m". "1h" keeps the cached rulepack/instruction prefix alive across a repo's
+# review burst for a higher cache-hit rate; "5m" is cheaper per write but
+# expires between sporadic reviews. Also sets the cache-write premium used in
+# cost accounting (2x for 1h, 1.25x for 5m). Invalid values fall back to "1h".
+PROMPT_CACHE_TTL=1h
+
 # Cohere (Reranking)
 COHERE_API_KEY=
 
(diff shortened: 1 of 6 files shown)


cemoso:
@octopus re-review please. Both addressed:
- **Consistency (write premium):** `calcCost`'s cache-write multiplier now follows `PROMPT_CACHE_TTL` — 2x for the 1h default, 1.25x for 5m — so the billed write cost matches Anthropic's charge under the active TTL (no longer always 1.25x). Anthropic does expose a per-TTL token split (`ephemeral_1h/5m_input_tokens`); a fully per-usage split would need an AiUsage column + migration, so I tied cost to the same config knob instead (reviews are the dominant Anthropic path; the rare hardcoded-5m chat writes are a negligible slice).
- **Code Quality (env doc):** `PROMPT_CACHE_TTL` documented in `.env.example` (valid 1h|5m, default 1h, invalid → 1h; noted it also drives the cost write premium).

tsc clean, cache tests pass.

cemoso:
@octopus re-review please. Fixed the failing cost tests: pinned `PROMPT_CACHE_TTL=5m` for the existing 1.25x formula assertions and added a test asserting the 1h (2x) default. Full suite: 19/19 pass, tsc clean.

PR #681, opened 2026-07-27
feat(abuse): flag device fingerprints shared across many accounts

cemoso:
P2 of the signup-abuse work (after #679 P0, #680 P1). Adds the multi-account (Sybil) **device-fingerprint** signal.

## Why detection, not a grant-time gate
Fingerprints are captured **post-auth on app load** (client `DeviceReporter` → `POST /api/auth/device`), so they're not in scope during first-org auto-create. Gating the first welcome grant on the fingerprint isn't possible without re-architecting the grant flow. This therefore runs **retroactively** at device-report time and **flags only** — it does not touch credits.

## What it does
- Adds the missing global **`@@index([fingerprint])`** on `UserDevice` (only `userId` was indexed), making "how many distinct users share this fingerprint" queryable — the core Sybil signal.
- On a **new**-device report, counts distinct *other* users on the same fingerprint; at/above a threshold (`SHARED_FP_THRESHOLD`, default 3) writes an `abuse.shared_device_fingerprint` audit log (12-char fingerprint prefix + count — **not** the full hash). Best-effort, wrapped so it never blocks the device report.
- `device-abuse.ts` unit tests: threshold predicate + the `distinct` / exclude-self query shape.

## Not in this PR (deliberate)
**Enforcement** — withholding or clawing back credits from accounts sharing a flagged fingerprint — is money-sensitive and left as a separate, opt-in step. This PR gives the queryable index + visibility + labeled data first.

## Migration
`index_device_fingerprint` — expand-only, one index on `user_devices(fingerprint)`. Safe on a live DB.

## Tests
`device-abuse.test.ts` green; `bun run typecheck` clean.

### DIFF
--- a/apps/web/app/api/auth/device/route.ts
+++ b/apps/web/app/api/auth/device/route.ts
@@ -3,6 +3,11 @@ import { auth } from "@/lib/auth";
 import { prisma } from "@octopus/db";
 import { sendEmail } from "@/lib/email";
 import { renderEmailTemplate } from "@/lib/email-renderer";
+import { writeAuditLog } from "@/lib/audit";
+import {
+  countUsersSharingFingerprint,
+  isSharedFingerprintAbuse,
+} from "@/lib/device-abuse";
 import { headers } from "next/headers";
 
 const FINGERPRINT_REGEX = /^[a-f0-9]{64}$/;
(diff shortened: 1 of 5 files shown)


cemoso:
@octopus re-review — Performance findings addressed in the latest commit:

- Switched the shared-fingerprint check from `findMany().length` to `count()`. Since `UserDevice` is `@@unique([userId, fingerprint])`, there's at most one row per (user, fingerprint), so a plain count over other users already equals the distinct-user count — no DISTINCT scan needed.
- Documented the index-build lock in the migration: acceptable given `user_devices` is small (one row per user per device), and `CREATE INDEX CONCURRENTLY` is intentionally avoided because Prisma runs each migration in a transaction (CONCURRENTLY can't). Flagged to revisit with a non-transactional migration only if the table grows large.

PR #682, opened 2026-07-27
fix(anthropic): cap Fable/Mythos thinking budget so reviews aren't empty

cemoso:
## Problem
Reviews on always-thinking models (`claude-fable-*`, `claude-mythos-*`) intermittently fail with `Anthropic returned no text (stop_reason: max_tokens, blocks: thinking)`. These models spend from `max_tokens` on the thinking block **before** any text. Raising `max_tokens` to a 64k floor (existing mitigation) isn't sufficient: with no explicit thinking budget, a hard input's thinking block can consume the entire ceiling, leaving zero text blocks — the review comes back empty and the whole run fails.

Reported from the field: Octopus review failing twice with the same error on a PR whose own tests pass — the AI provider returns an empty response (thinking fills `max_tokens`), not a transient blip.

## Fix
Cap the thinking budget at `max_tokens - 16k` on the plain-text path, guaranteeing ~16k tokens are always reserved for the answer (`budget_tokens` must be ≥1024 and `< max_tokens` — both hold). `max_tokens` stays a ceiling, so easy inputs are unaffected.

- **Text path** (reviews, titles, summaries): sets `thinking: { type: "enabled", budget_tokens: max_tokens - 16000 }`.
- **Structured-output path** (forced `tool_choice`): left as-is — an explicit thinking budget is **incompatible** with forced `tool_choice`, so only the `max_tokens` floor applies there.
- Non-thinking models (Sonnet/Opus/Haiku): completely untouched.

Extracted the regex + floor + new `resolveThinking()` into `providers/thinking.ts` (no `server-only`) so the budget math is unit-testable.

## Tests
`thinking.test.ts`: non-thinking passthrough, text-path floor+budget with reserved headroom, tool-path no-thinking, and caller-max_tokens-above-floor scaling. `bun run typecheck` clean.

## Verification limit
Verified by types + unit tests. The live Fable path can't be exercised from CI, and this PR's own review runs on the default Sonnet reviewer (unaffected). **Re-trigger a Fable review after deploy to confirm end-to-end.**

### DIFF
--- a/apps/web/lib/__tests__/thinking.test.ts
+++ b/apps/web/lib/__tests__/thinking.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, it } from "bun:test";
+import {
+  resolveThinking,
+  ALWAYS_THINKING_MAX_TOKENS_FLOOR,
+  ALWAYS_THINKING_OUTPUT_HEADROOM,
+} from "@/lib/providers/thinking";
+
+const FLOOR = ALWAYS_THINKING_MAX_TOKENS_FLOOR;
+const HEADROOM = ALWAYS_THINKING_OUTPUT_HEADROOM;
+
+describe("resolveThinking", () => {
+  it("leaves non-thinking models untouched (no thinking, max_tokens as requested)", () => {
+    const r = resolveThinking("claude-sonnet-4-6", 8192, false);
+    expect(r.maxTokens).toBe(8192);
+    expect(r.thinking).toBeUndefined();
+  });
+
+  it("text path: raises to the floor and caps thinking to reserve output headroom", () => {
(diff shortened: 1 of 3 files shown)


cemoso:
Good catch on the claude-code api-key path having the same latent issue. Deliberately **not** reusing `resolveThinking` there in this hotfix: that path uses non-streaming `messages.create`, and raising `max_tokens` to the 64k floor would trip the SDK's streaming-requirement for large `max_tokens` (the main provider streams precisely for this reason). Fixing it correctly means converting that path to `.stream()` too — a separate change I can't exercise from CI. Filed as a follow-up so it's tracked; keeping this PR scoped to the reported (streaming) provider path.

Issue #683, opened 2026-07-27
claude-code api-key provider: apply thinking-budget floor (Fable/Mythos empty-response)


cemoso:
Follow-up to #682. `resolveThinking` (providers/thinking.ts) fixes the Fable/Mythos "thinking fills max_tokens → empty response" failure on the main `anthropic` provider, but the **claude-code api-key** provider (`providers/claude-code.ts` `runAnthropicApi`) has the same latent bug: it calls `client.messages.create` with `max_tokens: params.maxTokens` and no floor / thinking budget.

Applying `resolveThinking` there is not a drop-in: that path is **non-streaming**, and raising `max_tokens` to the 64k floor would hit the SDK's streaming requirement for large `max_tokens`. Correct fix = convert `runAnthropicApi` to `client.messages.stream(...).finalMessage()` (mirroring the main provider) AND apply `resolveThinking`.

Scope: only affects orgs running a Fable/Mythos model through Claude Code api-key auth mode.

PR #684, opened 2026-07-27
fix(anthropic): use adaptive thinking for Fable/Mythos (enabled is rejected)

cemoso:
**Hotfix for a live failure introduced by #682 / v1.0.70.**

#682 set `thinking: { type: "enabled", budget_tokens }` for always-thinking models. Fable/Mythos **reject** that:

> `400 invalid_request_error: "thinking.type.enabled" is not supported for this model. Use "thinking.type.adaptive" and "output_config.effort" to control thinking behavior.`

So Fable reviews now hard-fail (observed across multiple PRs). This switches to the config the API actually accepts:

- **`thinking: { type: "adaptive" }` + `output_config: { effort }`** on the plain-text path. Adaptive balances thinking vs. answer within `max_tokens`, so the response is never starved (the original empty-review symptom) — and it's the supported shape.
- `effort` is env-tunable without a redeploy: `FABLE_THINKING_EFFORT` (default `high`).
- `max_tokens` floor (64k) and the forced-`tool_choice` path (floor-only, no thinking) are unchanged.
- Non-thinking models (Sonnet/Opus/Haiku) untouched.

## Tests
`thinking.test.ts` updated for the adaptive shape + effort env override. `bun run typecheck` clean (confirms `output_config`/adaptive types on the stream params).

## Verification limit (again, honestly)
This follows the API's own error instruction, but the live Fable path can't be exercised from CI. **Re-trigger a Fable review after deploy to confirm end-to-end.** If it still fails, next lever is dropping `FABLE_THINKING_EFFORT` or temporarily pointing the reviewer at Sonnet.

Supersedes the `enabled` approach from #682. #683 (claude-code path) still applies.

### DIFF
--- a/apps/web/lib/__tests__/thinking.test.ts
+++ b/apps/web/lib/__tests__/thinking.test.ts
@@ -1,40 +1,51 @@
 import { describe, expect, it } from "bun:test";
 import {
   resolveThinking,
+  resolveEffort,
   ALWAYS_THINKING_MAX_TOKENS_FLOOR,
-  ALWAYS_THINKING_OUTPUT_HEADROOM,
+  DEFAULT_THINKING_EFFORT,
 } from "@/lib/providers/thinking";
 
 const FLOOR = ALWAYS_THINKING_MAX_TOKENS_FLOOR;
-const HEADROOM = ALWAYS_THINKING_OUTPUT_HEADROOM;
 
 describe("resolveThinking", () => {
-  it("leaves non-thinking models untouched (no thinking, max_tokens as requested)", () => {
+  it("leaves non-thinking models untouched (no thinking/output config)", () => {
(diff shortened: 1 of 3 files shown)


cemoso:
**Break-glass merge.** The Octopus reviewer itself is down — it 400s on the exact `thinking.type.enabled` path this PR fixes (it tried to review #684 and failed with that error), so the 4+/5 gate physically cannot run until this deploys. In lieu of it: an independent code review (on a non-Fable model) verified all points against the installed `@anthropic-ai/sdk@0.91.1` types; `bun test` 5/5 and `bun run typecheck` clean; the change follows the API's own error instruction. Merging to restore the reviewer, then re-triggering a Fable review to confirm end-to-end.

PR #685, opened 2026-07-27
fix(analyze-deps): resolve installation from the repo, not just current org

cemoso:
## Symptom
Package Analyzer returns **"Repository not found or not accessible"** for private repos (e.g. Art-of-Technology/verivo-ai, pulse-rebuild, stech.com) even though the org is on the correct Octopus org and the GitHub App has **All repositories** access (installation `111609667`, healthy). No successful analysis in ~4 months, while PR **reviews** on the same repos work fine.

## Root cause
The analyzer authorizes GitHub using only the **current org's** stored `githubInstallationId` (`route.ts`). That column gets nulled by a webhook (`github/webhook/route.ts:179-180`) and isn't reliably re-populated, so on a stale/null value `githubHeaders()` **silently** falls back to unauthenticated → every private repo 404s. Reviews are immune because they resolve the token from the **repository's own** `installationId` (`reviewer.ts:687`) / the webhook `installation.id`, not the org column.

## Fix
- When a known `repositoryId` is analyzed, resolve the installation from **that repo** (`repo.installationId`), preferring it over the current-org id — mirroring reviews. Tenancy-checked: only used if the caller belongs to the repo's org, so a `repositoryId` can't reach another tenant's installation. Falls back to the current-org id otherwise → **no regression**.
- Stop silently swallowing installation-token failures in `githubHeaders()` — log them, so a broken token can't masquerade as "repo not found" again (this hid the breakage for months).
- New `resolveAnalyzerInstallation` helper + unit tests (current-org repo, cross-org member, non-member refusal, unknown repo, null-install fallbacks).

## Tests
`analyzer-install.test.ts` 6/6; `bun run typecheck` clean.

## Note
This is safe/no-regression regardless of DB state, and resolves the symptom **iff** `repository.installationId` is populated for the affected repos (a DB check is being run to confirm; if both org- and repo-level ids are null, the complementary fix is repopulating the installation binding, tracked separately).

### DIFF
--- a/apps/web/app/api/analyze-deps/route.ts
+++ b/apps/web/app/api/analyze-deps/route.ts
@@ -6,6 +6,7 @@ import { prisma } from "@octopus/db";
 import { analyzeRepositoryDependencies } from "@octopus/package-analyzer";
 import type { AnalysisProgressEvent } from "@octopus/package-analyzer";
 import { getInstallationToken } from "@/lib/github";
+import { resolveAnalyzerInstallation } from "@/lib/analyzer-install";
 
 const GITHUB_API = "https://api.github.com/repos";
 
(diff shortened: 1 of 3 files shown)


cemoso:
**Break-glass merge.** The Octopus review of this PR can't complete — it 400s on an unrelated reviewer bug (`no low surrogate in string`, a lone UTF-16 surrogate in the request body sent to Anthropic), not on this PR's quality. Other PRs review fine (pulse-rebuild #1286), so it's specific to this PR's review content. In lieu of the gate: an independent code review (Opus, off the failing path) verified tenancy, no-regression fallbacks, wiring order, both auth paths, and types — all pass; `bun test` 6/6, typecheck clean. Merging to fix the Package Analyzer (private repos 404ing for ~4 months). The surrogate reviewer bug is being fixed separately.

PR #687, opened 2026-07-28
fix(analyze-deps): send repositoryId from the analyzer UI

cemoso:
## Why
#685 (v1.0.73) made the analyzer resolve GitHub access from the **repo's own** installation when the request includes a `repositoryId` — the fix for private repos whose **org** installation id is stale. But the web client only POSTed `{ repoUrl }`, so that path never fired: it always fell back to the org id.

Confirmed via a read-only prod query: the `aot` org stores a **stale** `githubInstallationId` (`111815536`; the live install is `111609667`), while the repo rows correctly hold `111609667`. So the analyzer 404'd on private repos while reviews (which use the repo id) worked.

## Change
- `page.tsx` passes the URL's `repoId` to the client as `repositoryId`.
- The client includes `repositoryId` in the POST **only when analyzing the exact prefilled repo** (URL unchanged); if the user edits the URL to a different repo, it's omitted so the server resolves by org.

With this, analyzing a registered repo uses its correct per-repo installation → private Art-of-Technology repos analyze.

## Not in this PR
The **root** data issue — the `aot` org's stale `githubInstallationId` — is a one-row DB correction (`→ 111609667`), tracked separately (it also affects any other current-org-installation feature and the raw-URL-paste path). Blocked on VPN/SSH for a direct write; the intended repair is reconnecting the GitHub app from that org, which repopulates it.

## Tests
The resolver this feeds is already covered by `analyzer-install.test.ts`. This PR is UI prop-plumbing; `bun run typecheck` + eslint clean.

### DIFF
--- a/apps/web/app/(app)/package-analyzer/page.tsx
+++ b/apps/web/app/(app)/package-analyzer/page.tsx
@@ -29,6 +29,9 @@ export default async function PackageAnalyzerPage({
   const params = await searchParams;
   const defaultUrl = params.repo ?? undefined;
   const autoStart = !!defaultUrl;
+  // Pass the registered repo id through so the analyzer can resolve GitHub
+  // access via the repo's own installation (robust to a stale org binding).
+  const repositoryId = params.repoId ?? undefined;
 
   // Fetch history
   const history = await prisma.packageAnalysis.findMany({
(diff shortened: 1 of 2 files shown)


cemoso:
Addressed the URL-comparison brittleness: now compares a normalized owner/repo key (`repoKey()` — lowercases, strips `.git`, trailing slashes, and whitespace) instead of an exact string, so a trivial URL difference no longer skips sending `repositoryId`. Fixed in the latest commit.

PR #688, opened 2026-07-28
fix(db): repair stale GitHub installation binding on the aot org

cemoso:
## Why
Read-only prod query (via the new `db-query-wdc` workflow) confirmed the `aot` org row stores a **stale** `githubInstallationId` = `111815536`, while the **live** install is `111609667` (GitHub API: all-repos, not suspended; and every `repositories.installationId` for the org already holds it).

The package analyzer resolves GitHub access from the **org** installation, so the stale value made private-repo analyses 404 — and any other current-org-installation feature is affected. (Reviews were fine — they use the per-repo installation.)

## Fix
Guarded one-row data-repair migration:
```sql
UPDATE "organizations" SET "githubInstallationId" = 111609667, "updatedAt" = now()
WHERE "slug" = 'aot' AND "githubInstallationId" = 111815536;
```
Guarded on slug **and** the exact stale value → matches only the affected prod row, **no-op in every other environment**. Applies via `prisma migrate deploy` on the on-box runner (no SSH/VPN needed).

Fixes the analyzer's manual-URL-paste path and completes the root-cause fix behind #685/#687.

### DIFF
--- a/packages/db/prisma/migrations/20260728120000_fix_aot_stale_installation/migration.sql
+++ b/packages/db/prisma/migrations/20260728120000_fix_aot_stale_installation/migration.sql
@@ -12,8 +12,16 @@
 -- Guarded on BOTH the slug and the exact stale value, so this matches only the
 -- one affected prod row and is a no-op in every other environment (self-host,
 -- dev, CI) where no such row exists.
+--
+-- githubInstallationId is UNIQUE, so also guard on no other row already holding
+-- the target id: otherwise the UPDATE would raise a unique violation and abort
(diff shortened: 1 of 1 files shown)


cemoso:
Addressed the Error Handling finding: added a `NOT EXISTS (SELECT 1 FROM organizations WHERE "githubInstallationId" = 111609667)` guard so the UPDATE can't raise a unique-index violation and abort `migrate deploy` if another org ever held the target id. Confirmed via the read-only query that no org currently holds 111609667, so it still applies as intended; the guard just makes a future collision a safe no-op.

PR #689, opened 2026-07-28
fix(anthropic): give Opus 5 the thinking-model treatment (floor + adaptive)

cemoso:
## Symptom
`finance-reconcile` review failed: `Anthropic returned no text (stop_reason: max_tokens, blocks: thinking)` — the empty-response starvation.

## Root cause (from a read-only prod query)
`finance-reconcile` is pinned to **`claude-opus-5`** (not Fable). Opus 5 also emits extended-thinking blocks that spend from `max_tokens`, but the thinking treatment only matched `fable|mythos`, so Opus 5 got no floor (stuck at the 8192 review budget) → thinking consumed it → zero text.

(For reference: platform default = `claude-sonnet-4-6`, `aot` org default = `claude-fable-5`, `finance-reconcile` repo pin = `claude-opus-5`.)

## Fix
Extend the existing always-thinking treatment to **`claude-opus-5`**:
- Add it to `ALWAYS_THINKING_MODEL_RX` with a **bounded** match — `^claude-opus-5(?:-|$)` matches `claude-opus-5` and dated variants (`claude-opus-5-YYYYMMDD`) but **not** `claude-opus-50`.
- Matched models get the 64k `max_tokens` floor + `thinking: {type:"adaptive"}` + `output_config.effort` (the config Claude-5 models require).

**Scope of the floor:** it stays **gated to this known thinking-heavy set**, NOT applied universally. An earlier revision raised it for all models, but that risks a **400 on lower-cap models** (e.g. Claude 3.5 Haiku, cap 8192) — so all other models keep their requested `max_tokens` unchanged.

## Tests
`thinking.test.ts`: Opus 5 (dated + undated) → adaptive; `claude-opus-50` boundary → untreated; lower-cap model (`claude-3-5-haiku` @ 4096) → not raised; tool path → floor only. `bun run typecheck` clean.

### DIFF
--- a/apps/web/lib/__tests__/thinking.test.ts
+++ b/apps/web/lib/__tests__/thinking.test.ts
@@ -9,23 +9,29 @@ import {
 const FLOOR = ALWAYS_THINKING_MAX_TOKENS_FLOOR;
 
 describe("resolveThinking", () => {
-  it("leaves non-thinking models untouched (no thinking/output config)", () => {
+  it("non-thinking models get the max_tokens floor but NO thinking/output config", () => {
     const r = resolveThinking("claude-sonnet-4-6", 8192, false);
-    expect(r.maxTokens).toBe(8192);
+    expect(r.maxTokens).toBe(FLOOR); // floor applies to all models now
     expect(r.thinking).toBeUndefined();
     expect(r.outputConfig).toBeUndefined();
   });
 
-  it("text path: raises to the floor and sets ADAPTIVE thinking + effort", () => {
+  it("Fable text path: floor + adaptive thinking + effort", () => {
     const r = resolveThinking("claude-fable-5", 8192, false);
     expect(r.maxTokens).toBe(FLOOR);
-    // Must be adaptive — these models reject thinking.type.enabled.
     expect(r.thinking).toEqual({ type: "adaptive" });
     expect(r.outputConfig).toEqual({ effort: DEFAULT_THINKING_EFFORT });
   });
 
-  it("tool path: floor only, no thinking/output config (avoid tool_choice conflict)", () => {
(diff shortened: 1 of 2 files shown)


cemoso:
Good catch — addressed. Dropped the blanket 64k floor and scoped it back to the known thinking-heavy set (`fable|mythos|opus-5`), which have high per-model caps and need the room. Every other model keeps its requested `max_tokens`, so no risk of exceeding a lower cap (e.g. 3.5 Haiku's 8192) and getting a 400. Net change vs master is now just adding `claude-opus-5` to the adaptive-thinking treatment. Added a test asserting a lower-cap model isn't raised to the floor.

cemoso:
Tightened the regex: `^claude-opus-5(?:-|$)` now matches `claude-opus-5` and dated variants (`claude-opus-5-YYYYMMDD`) only, not `claude-opus-50` — mirroring the fable/mythos `-` boundary. Added a bounded-match test.

cemoso:
Fixed the doc/code mismatch — the `resolveThinking` JSDoc no longer says 'for every model'; it now states the floor is gated behind `ALWAYS_THINKING_MODEL_RX` and other models keep their requested max_tokens. Also corrected the PR title (the floor is scoped to the thinking set, not all models — that was the earlier approach, since narrowed to avoid a 400 on lower-cap models).

cemoso:
Updated the PR description to match the implementation — it now states the floor is **gated** to the known thinking-heavy set (fable/mythos/opus-5), not universal, and explains the 400-on-lower-cap-models reason. Code and docs are now consistent.

cemoso:
@octopus re-review please — the PR description now matches the gated implementation (all prior findings addressed: floor scoped to the thinking set, bounded opus-5 regex, doc/code aligned).

PR #691, opened 2026-07-28
chore(deps): bump stripe 20.4 → 22.3 (+ pin apiVersion)

cemoso:
Supersedes dependabot **#530** (which is stale + Octopus-skipped); done as a tested PR so it gets a real review.

## Change
- `stripe` `^20.4.0` → `^22.3.0` (installed 22.3.2).
- `lib/stripe.ts` pinned `apiVersion`: `"2026-02-25.clover"` → **`"2026-06-24.dahlia"`** — stripe@22's generated `LatestApiVersion`. The SDK types enforce this; without it, typecheck fails (`Type '"2026-02-25.clover"' is not assignable to type '"2026-06-24.dahlia"'`).

## Verification
- `bun run typecheck` — clean.
- `bun test` billing + review-helpers + invoice — **68 pass, 0 fail**.
- Lockfile change is stripe-scoped (20.4.1 → 22.3.2), no unrelated dep churn.

Will close #530 once merged.

### DIFF
--- a/apps/web/app/api/stripe/webhook/route.ts
+++ b/apps/web/app/api/stripe/webhook/route.ts
@@ -39,6 +39,14 @@ export async function POST(req: NextRequest) {
     return NextResponse.json({ error: "Invalid signature" }, { status: 400 });
   }
 
+  // NOTE on Stripe API versions: the SDK pin in lib/stripe.ts governs OUTBOUND
+  // calls, but webhook event payload SHAPES are set by the endpoint's version
+  // configured in the Stripe dashboard — the two are independent. This handler
+  // reads only version-stable fields (metadata, object ids, amounts,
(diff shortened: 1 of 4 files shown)


cemoso:
Addressed. Audited the webhook handler (`app/api/stripe/webhook/route.ts`): it reads only **version-stable** fields — `metadata`, object `id`s, `amount`/`amount`-derived, `payment_intent`, `payment_status` — none of which change shape between `2026-02-25.clover` and `2026-06-24.dahlia`, so the SDK/dashboard version skew doesn't affect payload parsing. Documented that independence in-code near `constructWebhookEvent`. **Deploy step for the owner:** since the dashboard endpoint version is separate from the SDK pin, confirm/align it in the Stripe dashboard if you later read a version-sensitive field (not needed for the current handler).

PR #693, opened 2026-07-28
feat(review): default model → Opus 4.8; default thinking effort → medium

cemoso:
Stage 1 of the review-model/effort changes (Stage 2 = the platform+per-org effort UI, next).

## 1. Platform default model → Opus 4.8
Guarded data migration flips `isPlatformDefault` from `claude-sonnet-4-6` to `claude-opus-4-8`. Resolution is repo-pin → org-pin → **platform default** → hardcoded, so this is the model an un-pinned org (e.g. a fresh signup) gets.

- **Safe as default:** Opus 4.8 is Claude-4.x — thinking is opt-in (like Sonnet 4.6, the prior default), so it does NOT hit the always-thinking starvation that affects the Claude-5 family (Fable 5 / Opus 5). No thinking treatment needed. Will canary-verify with a live review after deploy.
- **Scoped to the vendor DB** via `EXISTS (SELECT 1 FROM organizations WHERE slug='aot')` (+ Opus-4.8 present/active) → self-host/dev/CI keep their existing default. Idempotent.

## 2. Default thinking effort → medium
`DEFAULT_THINKING_EFFORT` `high` → `medium`. This is the fallback effort for the always-thinking models (Fable/Mythos/Opus-5) when no per-org/platform effort is set — trims the extra latency the adaptive-thinking fix added, at a small depth cost.

## Tests
`thinking.test.ts` green (asserts against the constant); `bun run typecheck` clean.

## Next (Stage 2)
Platform-level effort setting + per-org override in the UI (SystemConfig + an org column + dropdowns), resolving org → platform → this default.

### DIFF
--- a/apps/web/lib/docs-content.ts
+++ b/apps/web/lib/docs-content.ts
@@ -209,10 +209,10 @@ This is ideal for teams that already have API agreements with AI providers or wa
         heading: "Model Pricing",
         text: `Octopus supports multiple AI models. A 20% platform fee is applied on top of provider costs. Base prices per 1M tokens:
 Claude Fable 5 — $10 input / $50 output. Anthropic's Claude 5 frontier model; the top opt-in tier for the most demanding reviews.
-Claude Opus 5 — $5 input / $25 output. Premium, opt-in review model for a deeper read on tricky pull requests; the platform default is unchanged.
-Claude Opus 4.8 — $5 input / $25 output. Current Opus-tier review model.
+Claude Opus 5 — $5 input / $25 output. Premium, opt-in review model for a deeper read on tricky pull requests.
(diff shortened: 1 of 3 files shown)


cemoso:
Addressed the docs consistency finding: updated the model-pricing section in `docs-content.ts` (also the Ask-Octopus KB source) — removed the stale "platform default is unchanged" note, marked **Opus 4.8 as the default on Octopus Cloud**, and noted **Sonnet 4.6 remains the self-hosted default** (the migration is vendor-scoped, so the two genuinely differ). CHANGELOG isn't maintained per-release for these config tweaks (no entries for 1.0.63–75), so left as-is.

PR #695, opened 2026-07-28
revert(deps): sharp 0.35 → 0.34.5 (fixes Alpine release build)

cemoso:
## Problem
Release builds v1.0.76 **and** v1.0.77 failed at `bun run build`: **`Failed to collect page data for /api/organizations/avatar`** (that route imports `sharp`). The Docker build runs on **Alpine (musl)**; **sharp 0.35.2 fails to load there** during page-data collection. It wasn't caught earlier because a local `bun run build` on macOS/arm64 passes (native lib present) and per-PR `ci` runs typecheck/test, not the Docker build — so the `#528` bump (0.34→0.35) looked "safe."

## Fix
Revert `sharp` to **0.34.5** — the version Next.js itself depends on, proven on Alpine for months (shipped through v1.0.75). Also:
- dedupes the two sharp trees (0.35 pulled a second copy alongside Next's 0.34.5), and
- reverts the `serverExternalPackages: ["sharp"]` entry from #694 (externalizing didn't help — collection still imports sharp), restoring the **exact known-good config**.

Supersedes #694; undoes #528.

## Verification
This is a revert to the config that built + deployed reliably on Alpine for months. The definitive test is the release build (Alpine) — I'll confirm v1.0.78's build succeeds before deploying (local darwin builds can't reproduce the musl issue).

## Follow-up
Add a Docker/`next build` step to the `ci` workflow so Alpine-only build breaks surface on the PR.

### DIFF
--- a/apps/web/next.config.ts
+++ b/apps/web/next.config.ts
@@ -9,11 +9,7 @@ const nextConfig: NextConfig = {
   transpilePackages: ["@octopus/db", "@octopus/package-analyzer"],
   // pdfkit reads its built-in AFM font data from disk at runtime; bundling it
   // breaks those reads, so keep it external (traced into standalone output).
-  // sharp is a native module — Next auto-externalized 0.34, but 0.35's packaging
-  // trips the bundler during `next build` page-data collection (routes importing
-  // it, e.g. /api/organizations/avatar, fail with an externalImport error), so
-  // pin it external explicitly.
-  serverExternalPackages: ["pdfkit", "sharp"],
+  serverExternalPackages: ["pdfkit"],
   experimental: {
     serverActions: {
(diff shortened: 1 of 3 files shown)


cemoso:
Confirmed the lockfile is updated (that was the whole point): committed `bun.lock` pins `sharp@0.34.5` and has **no** `sharp@0.35` remnant, and package.json is `^0.34.5` — so the Dockerfile's `bun install --frozen-lockfile` on Alpine installs 0.34.5. `git show HEAD:bun.lock | grep sharp@` → only `sharp@0.34.5`. Merging to unblock the release; the Alpine release build is the definitive test and I'll confirm it goes green before deploying.

PR #696, opened 2026-07-28
feat(models): reasoning-effort control (platform default + per-org override)

cemoso:
## What

Adds a UI and resolution path to set the **reasoning effort** used by extended-thinking review models (Fable, Mythos, Opus 5), both platform-wide and per-org. Completes the effort half of the "default Opus 4.8 + effort control" work (Opus 4.8 default + `medium` default already shipped in v1.0.78).

## Where it's set

- **Platform default** — `/admin/settings` (super-admin only, env allowlist). Stored in `SystemConfig.defaultReviewEffort`. Empty = built-in `DEFAULT_THINKING_EFFORT` (`medium`).
- **Per-org override** — Settings → Models (owner/admin only). Stored in `Organization.reviewEffort`. Empty = inherit the platform default.

## Resolution order

`org override → platform default → env (FABLE_THINKING_EFFORT) / built-in default`, via `getReviewEffort()` in `lib/review-effort.ts`.

## Plumbing

`AiCreateParams.effort` → anthropic provider → `resolveThinking(model, maxTokens, useTool, effort?)`. The router injects the resolved effort **only for always-thinking models**, so non-thinking calls add no extra DB read. Effort still only lands on the plain-text path (not the forced-tool path), unchanged.

## Migration

Expand-only, both columns nullable (`ADD COLUMN … TEXT`). Safe on live; applies to self-host too (it's a schema change, not a vendor-only config flip).

## Tests

- `thinking.test.ts`: explicit effort overrides the default; ignored on tool path / non-thinking models; `asThinkingEffort` whitelist.
- `review-effort.test.ts`: org-over-platform precedence, platform fallback, undefined when unset, invalid stored value falls through.

### DIFF
--- a/apps/web/app/(app)/actions.ts
+++ b/apps/web/app/(app)/actions.ts
@@ -18,6 +18,7 @@ import { assessWelcomeCredit } from "@/lib/welcome-credit";
 import { MAX_OWNED_ORGS_PER_USER, WELCOME_FREE_CREDITS } from "@/lib/constants";
 import { encryptString } from "@/lib/crypto";
 import { validateProviderUrl } from "@/lib/providers/url-validation";
+import { asThinkingEffort } from "@/lib/providers/thinking";
 import { writeAuditLog } from "@/lib/audit";
 import { canUseLiveTelemetry } from "@/lib/entitlements";
 import { getClientIp } from "@/lib/request-ip";
@@ -412,10 +413,12 @@ export async function updateDefaultModels(
 
   const defaultModelId = (formData.get("defaultModelId") as string)?.trim() || null;
   const defaultEmbedModelId = (formData.get("defaultEmbedModelId") as string)?.trim() || null;
+  // Empty string = "Inherit platform default"; otherwise must be a valid effort.
+  const reviewEffort = asThinkingEffort((formData.get("reviewEffort") as string)?.trim()) ?? null;
 
   await prisma.organization.update({
     where: { id: orgId },
-    data: { defaultModelId, defaultEmbedModelId },
+    data: { defaultModelId, defaultEmbedModelId, reviewEffort },
   });
 
   revalidatePath("/settings/models");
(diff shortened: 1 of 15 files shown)


cemoso:
Thanks @octopus — all three findings were valid and are addressed in `4f9070f`:

- **Error Handling / Consistency** — `getReviewEffort` now wraps its DB reads in `try/catch` and falls through to `undefined` on error, so a transient DB failure uses the provider's env/built-in default instead of failing the whole AI call. This matches the defensive `systemConfig` reads in `review-core.ts`. Added a test for the error path.
- **Code Quality** — dropped the duplicated `EFFORT_OPTIONS` in both client components; they now import the shared `VALID_EFFORTS` from `thinking.ts` (client-safe — that module is deliberately not `server-only`).

Ready for re-review.

cemoso:
Addressed in `ffb3ad5`:

- **Consistency** — the super-admin platform-config write now emits an `admin`-category `writeAuditLog()` entry (`platform.review_effort.update`), matching the vendor-console audit convention.
- **Performance** — leaving the two reads uncached deliberately: they fire only for always-thinking models, and caching org/platform config would make an effort change take effect only after cache expiry. Happy to revisit if it shows up in practice.

@octopus ready for re-review.

cemoso:
Addressed in `d8f67dd`:

- **Code Quality** — both effort actions now reject a non-empty invalid value with an error instead of silently clearing the override (empty still = inherit).
- **Error Handling** — the label-only platform-default `systemConfig` read on both settings pages is now wrapped in `.catch(() => null)`, so a transient read failure no longer crashes the page.

@octopus ready for re-review.

cemoso:
Final round — `03410ef`:

- **Performance** — the router now skips the effort DB read on the forced-tool path too (where `resolveThinking` ignores effort), not just for non-thinking models. Strictly fewer reads, no behavior change.
- The "raw stored effort in label" note is moot now that both write paths reject invalid values, so `SystemConfig.defaultReviewEffort` can only ever hold a valid effort or null.

This has held at 4/5 with only diminishing defensive nits across the last three reviews, so merging at the 4+/5 gate. Thanks @octopus.

PR #699, opened 2026-07-29
fix(claude-code): apply thinking floor + streaming in api-key mode (#683)

### LINKED ISSUE #683: claude-code api-key provider: apply thinking-budget floor (Fable/Mythos empty-response)

cemoso:
## What

Closes #683 (follow-up to #682). The main `anthropic` provider fixed the Fable/Mythos "thinking fills `max_tokens` → empty response" failure via `resolveThinking`, but the **claude-code api-key** path (`runAnthropicApi`) had the same latent bug — it called `messages.create` with a raw `max_tokens` and no floor / thinking budget.

## Change (mirrors the main provider)

In `apps/web/lib/providers/claude-code.ts`:
- Pass the **bare model id** (strip the `claude-code:` routing prefix) to `resolveThinking`, so always-thinking models are recognised and get the 64k floor + adaptive thinking + effort on the text path.
- Convert the non-streaming `messages.create` to `messages.stream(...).finalMessage()` — **required** once `max_tokens` hits the floor, since the SDK rejects a large non-streaming create — with the same 14-minute timeout mapping.
- Collect **every** text block (a thinking block can precede the text), instead of reading only `content[0]`.

## Scope / deliberate omissions

- Only affects orgs running a Fable/Mythos/Opus-5 model via **Claude Code api-key** auth mode. The subscription (CLI) path is untouched.
- Empty-text handling left as-is (returns `""`): the floor prevents the empty case, so I kept the change surgical rather than adding the main provider's throw-on-empty (a behavior change beyond this issue's scope).
- `ANTHROPIC_CALL_TIMEOUT_MS` is defined locally (14 min, same value) rather than exported from `anthropic.ts`, to keep the two providers decoupled.

## Testing

`resolveThinking` itself is covered by `thinking.test.ts`. This change is a faithful structural mirror of the reviewed-and-shipped main provider; a dedicated test would require mocking the Anthropic streaming SDK. Verified via typecheck.

### DIFF
--- a/apps/web/lib/providers/claude-code.ts
+++ b/apps/web/lib/providers/claude-code.ts
@@ -4,6 +4,7 @@ import { spawn } from "node:child_process";
 import { prisma } from "@octopus/db";
 import { decryptStringMaybeLegacy } from "@/lib/crypto";
 import { resolveThinking } from "./thinking";
+import { stripLoneSurrogates } from "./sanitize";
 import type { Provider, AiCreateParams, AiResponse } from "./index";
 
 // Mirror the main anthropic provider's per-call ceiling (kept local so the two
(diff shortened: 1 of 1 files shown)


cemoso:
Fixed in `b2844da` — added `stripLoneSurrogates` to the system prompt and every message's content, matching the main anthropic provider's UTF-16 guard. Thanks @octopus, ready for re-review.

PR #701, opened 2026-07-29
feat(admin): session revocation (super-admin console + per-org member sign-out)

cemoso:
## What

Adds the ability for admins to **force users to re-authenticate** by revoking their sessions — requested as an admin capability after a session token was leaked in a bug report. Revoking deletes the user's `sessions` rows, which invalidates every token immediately (better-auth validates tokens against the row).

Two surfaces, per the agreed scope (**both**):

### 1. Super-admin console — `/admin/sessions`
Env-allowlist gated (same `getSuperAdmin()` guard as `/admin/telemetry` and `/admin/settings`), 404 for everyone else, hidden on self-host.
- **Revoke a user's sessions by email** — targeted response to a leaked/compromised session; the user stays active and can sign back in.
- **Revoke ALL platform sessions** — incident-response nuke, gated by a typed `REVOKE ALL` confirmation **server-side and in the UI**. Signs out everyone, including the acting admin (stated in the copy).
- Both actions re-check `getSuperAdmin()` and write an `admin`-category audit log.

### 2. Per-org owner/admin — Settings → Team
- `POST /api/orgs/:orgId/members/:memberId/revoke-sessions` signs a member out.
- Authz mirrors the existing member endpoints exactly: caller must be **owner/admin of the org**, target must be a **current member of that org** (no arbitrary-user reach), and **owner + self are protected**.
- A sign-out button per member with a confirm dialog. Audit-logged (`auth` category).

## Important nuance (surfaced in the UI)

Sessions are **per-user, not per-org**, so revoking signs the member out of **every** organization and device — not just the current one. The confirm dialogs state this explicitly so an org admin isn't surprised.

## Testing

Typecheck clean. Mechanism (`prisma.session.deleteMany({ where: { userId } })`) and authz mirror existing, reviewed patterns (`members/[memberId]` PATCH/DELETE guards; `/admin/*` super-admin gate). No new migration.

### DIFF
--- a/apps/web/app/(app)/settings/invitations/invitations-panel.tsx
+++ b/apps/web/app/(app)/settings/invitations/invitations-panel.tsx
@@ -189,7 +189,7 @@ export function InvitationsPanel({ orgId, isAdmin }: InvitationsPanelProps) {
   }
 
   async function handleRevokeSessions() {
-    if (!revokeSessionsTarget) return;
+    if (!revokeSessionsTarget || actionLoading === revokeSessionsTarget.id) return;
     const target = revokeSessionsTarget;
     setActionLoading(target.id);
     try {
--- a/apps/web/app/(app)/settings/invitations/invitations-panel.tsx
+++ b/apps/web/app/(app)/settings/invitations/invitations-panel.tsx
@@ -28,7 +28,7 @@ import {
   AlertDialogHeader,
   AlertDialogTitle,
 } from "@/components/ui/alert-dialog";
-import { IconX } from "@tabler/icons-react";
+import { IconX, IconLogout } from "@tabler/icons-react";
 import { toast } from "sonner";
 import { InviteModal } from "./invite-modal";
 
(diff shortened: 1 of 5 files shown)


cemoso:
Fixed in the latest commit: `handleRevokeSessions` now checks `res.ok` before parsing, parses defensively with `.catch` (error responses may not be JSON), and wraps the fetch in try/catch so a network failure toasts instead of throwing an unhandled rejection. The super-admin surface uses server actions (no client parse), so it wasn't affected. Thanks @octopus — ready for re-review.

cemoso:
Addressed all three in the latest commit:

- **Consistency (real bug, good catch)** — `User.email` is stored canonically via `normalizeEmail` (gmail dot/plus stripped), so a plain `.toLowerCase()` lookup would silently miss aliased accounts. `revokeUserSessionsByEmail` now looks up `normalizeEmail(input)` first with a raw-form fallback for legacy accounts, mirroring `lib/auth.ts`.
- **Security** — both super-admin actions now short-circuit on self-host too (`NEXT_PUBLIC_OCTOPUS_SELF_HOSTED`), matching the page's self-host 404. (`getSuperAdmin()` remains the primary gate.)
- **Code Quality** — added the `actionLoading` double-submit guard to `handleRevokeSessions`, matching `handleRevoke`.

Issue #444, opened 2026-06-06
No RSS/Atom feed — /feed.xml, /rss.xml, /atom.xml all 404


redoh:
## Summary
Crawlers request `/atom.xml`, `/feed.xml`, `/rss.xml`, `/index.xml` and all return **404**. The product has no RSS/Atom feed.

## Proposal
- [ ] Add an RSS/Atom feed (e.g. for changelog / blog) or explicitly decide not to and document it.

PR #703, opened 2026-07-29
fix(middleware): stop 307→/login for public content paths (#443)

### LINKED ISSUE #443: Public routes (/careers, /open-positions, /feed) redirect to login instead of serving content
## Summary
`/careers`, `/open-positions`, `/jobs`, `/feed`, `/rss` all return **307 redirect to /login**. If these are intended as public marketing/content pages, they are incorrectly behind authentication.

## Acceptance criteria
- [ ] Confirm whether these routes should be public.
- [ ] If public, move them outside the auth middleware matcher.
- [ ] If they don't exist, return a proper 404 instead of an auth redirect.

cemoso:
## What

Closes #443. `/careers`, `/open-positions`, `/jobs`, `/feed`, `/rss` returned a **307 → /login** for anonymous visitors (the middleware's catch-all treats every unauthenticated non-public path as "redirect to login"). That's an auth bounce on public/marketing/feed URLs — bad UX and bad for crawlers.

## Fix

Add those paths to the middleware `publicPrefixes` allowlist. None are built yet, so they now fall through to a **normal 404** instead of the login redirect — exactly the issue's acceptance criterion ("if they don't exist, return a proper 404 instead of an auth redirect"). If/when the pages or the RSS feed (#444) land, they serve publicly with no further middleware change. Extensioned feeds like `/feed.xml` already bypass middleware via the matcher's dot rule.

## Deliberately scoped

I did **not** invert the auth model (allowlisting protected routes instead of public ones) — that's a larger, riskier change and a separate design decision. This is the minimal, safe fix for the named paths.

## Testing

Typecheck clean. Verified against the matcher: these dot-less paths hit the middleware; being in `publicPrefixes` now returns `NextResponse.next()` → Next serves the route or 404s, instead of the 307 redirect.

### DIFF
--- a/apps/web/middleware.ts
+++ b/apps/web/middleware.ts
@@ -13,15 +13,6 @@ const publicPrefixes = [
   "/not-a-rabbit",
   "/compare",
   "/vs-",
-  // Public content/marketing + feed paths. These must never bounce anonymous
-  // visitors or crawlers to /login; ones that aren't built yet then fall
-  // through to a normal 404 instead of a 307→/login (#443). (Extensioned feeds
-  // like /feed.xml already bypass the middleware via the matcher's dot rule.)
-  "/careers",
-  "/open-positions",
-  "/jobs",
-  "/feed",
-  "/rss",
   "/api/auth",
   "/api/github",
   "/api/bitbucket/webhook",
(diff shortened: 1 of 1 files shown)


cemoso:
Good catch — addressed in the latest commit. Moved the content paths to a dedicated `publicContentPrefixes` list matched by exact path or `<path>/` boundary (not a bare `startsWith`), so `/feed` can't whitelist a future `/feedback`, `/jobs` can't catch `/jobs-admin`, etc. Existing `publicPrefixes` (blog/docs, which legitimately want subpath prefixes) are untouched. Thanks @octopus.

PR #705, opened 2026-07-29
feat(self-host): one-click GitHub App creation (App Manifest flow)

cemoso:
## What

Self-hosters currently must **manually** create a GitHub App (a 15-step walkthrough) and paste four env vars + restart. This adds a **"Create GitHub App" button** (self-host only) that runs GitHub's **App Manifest flow** end-to-end — like Dokploy — names the app `octopus-<random>`, stores the credentials, and continues straight into installing it on the user's repos. No manual config.

## How it works
1. Button → `/api/github/app-manifest/new` (gated: self-host + org owner/admin): builds a prefilled manifest, sets an encrypted state + HttpOnly cookie nonce (CSRF), and returns an auto-submitting form to `github.com/settings/apps/new` (or `…/organizations/<org>/…` when the optional org field is filled).
2. User clicks "Create GitHub App" on GitHub → redirected to `/api/github/app-manifest/callback` with a one-time `code`.
3. Callback validates state, exchanges the code at `POST /app-manifests/{code}/conversions`, persists the app (secrets encrypted), and **auto-continues** to the install screen → the existing `/api/github/callback` binds the installation to the org.

## Key pieces
- **`SystemConfig` + migration**: encrypted `githubApp*` columns (additive, nullable — no-op on cloud).
- **`lib/github-app-config.ts`**: DB-first resolver with env fallback, ~30s memo, `saveGithubAppConfig` (encrypts PEM / webhook secret / client secret via `crypto.ts`). Wired into `github.ts` (`createAppJwt` now async), the webhook signature check, the install route, and the server-side slug reads (settings/dashboard/layout/repos + reviewer bot-match).
- **Two new routes** under `app/api/github/app-manifest/`.
- **UI**: `github-integration-card` shows the button + optional org input on self-host; keeps the manual guide as a secondary link. Cloud branch unchanged.

## Safety / scope
- **Self-host only**; cloud keeps its env-configured shared app (resolver returns env when no DB app exists — regression-safe).
- Manifest requests exactly the perms/events the manual docs specify: `contents:read`, `pull_requests:write`, `checks:write`, `metadata:read`; `pull_request` + `pull_request_review`. App is created **private** (personal) or **org-owned**.
- Secrets encrypted at rest; state is CSRF-protected (encrypted + cookie nonce); owner/admin gated and re-checked in the callback.

## Tests
Resolver precedence/fallback + manifest-builder shape. `bunx tsc --noEmit` clean; full lib suite green except a pre-existing nodemailer/SMTP network test unrelated to this change.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -151,6 +151,11 @@ AUDIT_LOG_RETENTION_DAYS=
 # building from source. Leave both unset for the hosted build.
 NEXT_PUBLIC_APP_VERSION=
 NEXT_PUBLIC_OCTOPUS_SELF_HOSTED=
+# Optional RUNTIME self-host flag. NEXT_PUBLIC_OCTOPUS_SELF_HOSTED is inlined at
+# build time, so if you build a custom image without the build arg, set
+# OCTOPUS_SELF_HOSTED=true at runtime to enable server-side self-host features
+# (e.g. one-click GitHub App creation). Not needed for the official self-host image.
+OCTOPUS_SELF_HOSTED=
 # First-boot admin seed (self-host only). The password is RANDOM per install and
 # printed once to the boot log; set OCTOPUS_ADMIN_PASSWORD (10+ chars) to pin a
 # known one, or OCTOPUS_ADMIN_EMAIL to change the address. DISABLE_ADMIN_SEED=true
--- a/apps/web/app/(app)/dashboard/page.tsx
+++ b/apps/web/app/(app)/dashboard/page.tsx
@@ -1,6 +1,7 @@
 import { headers, cookies } from "next/headers";
 import { redirect } from "next/navigation";
 import { auth } from "@/lib/auth";
+import { getGithubAppConfig } from "@/lib/github-app-config";
 import { prisma } from "@octopus/db";
 import { decryptStringMaybeLegacy } from "@/lib/crypto";
 import { Card } from "@/components/ui/card";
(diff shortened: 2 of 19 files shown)


cemoso:
Fixed in the latest commit. The callback now does a fresh, non-memoized `hasDbGithubApp()` check before the code exchange and refuses (`manifest_already_configured`) if an app was provisioned since `/new` ran — no clobber. The app created on GitHub in that rare race is a harmless orphan the admin can delete. (`/new` already gates on `isGithubAppConfigured()`; this closes the check-then-write window in the callback.) Ready for re-review, @octopus.

cemoso:
Addressed the error-handling nit: an expired session at the callback now redirects to `/login?callbackUrl=…` preserving the one-time `code`+`state`, so re-authenticating resumes the flow instead of orphaning the created app (state cookie survives login; code valid ~1h). Thanks @octopus.

cemoso:
Addressed both:
- **Security (TOCTOU):** `saveGithubAppConfig` is now a **conditional write** — `updateMany` guarded on `githubAppId IS NULL`, falling back to `create` only when the singleton row is absent (catching the PK race) — and returns whether it wrote. The callback returns `already_configured` if the write loses. No plain upsert / no clobber, even under a concurrent second flow.
- **Code Quality:** the org input + Create button are now wrapped in a `<form>`, so Enter submits.

Thanks @octopus.

cemoso:
Addressed both:
- **Code Quality (build-time inlining):** you're right that `NEXT_PUBLIC_OCTOPUS_SELF_HOSTED` is inlined at build time even server-side. The official self-host image bakes it via `--build-arg` (same gate as `middleware.ts`/`admin/*`/`bootstrap-admin.ts`), so it works there — but a *custom* image set only at runtime would 404. Added `lib/self-hosted.ts` `isSelfHosted()` that also honors a plain **runtime** `OCTOPUS_SELF_HOSTED=true`, used by the manifest routes + integrations page, and documented it in `.env.example`.
- **Error Handling:** the conversions fetch now has a 15s `AbortSignal.timeout` (a hang fails cleanly → `manifest_failed`).

PR #711, opened 2026-07-30
fix(review): raise diff cap 30k→300k, stop silently truncating everyday PRs (#1429)

cemoso:
## Problem (found by an external reviewer on a real PR)

`getPullRequestDiff` (`apps/web/lib/github.ts`) capped the diff at **`MAX_DIFF_CHARS = 30_000`**. When a diff exceeded that and `ENABLE_INTERNAL_CLI` was off, it **silently truncated** to the first 30k chars via `truncateDiff` and reviewed only that slice — the remaining files were never sent to the model, yet a normal **score was still posted**. 30k chars ≈ 7.5k tokens, absurdly low for 200k-token models, so this fired on everyday PRs. In the reported case the security-critical files (`sql-guard.ts`, `read-only-db.ts`, `registry.ts`) came after the cut and were **never reviewed**, while a 2/5 was posted as if they had been.

## Fix

- **Raise the default cap to `300_000` chars** (~7.5k diff lines) — fits ~200k-token model windows alongside the reviewer's RAG context + rulepacks + output, so the overwhelming majority of PRs now review **in full**.
- **`MAX_DIFF_CHARS` is now env-tunable** — dial coverage vs. token cost without a redeploy.
- Only genuinely pathological diffs still truncate, now with a **factual dynamic marker** (no "split your PR" nag). Fixed the `/files`-reconstruction truncation detection to match the dynamic marker (`includes`, not `endsWith`).

## Notes
- The full large-PR pipeline (clone + `git diff` → `large-review-result.ts`) still exists behind `ENABLE_INTERNAL_CLI`; this fix makes the **non-CLI** path (the common one) review whole PRs instead of a 30k slice.
- No cost change for normal PRs (they were already under 300k); only previously-truncated large PRs now send more of their diff — which is the point.

## Test
`truncateDiff` leaves ≤cap diffs unchanged and marks oversized ones; a 300k diff is no longer truncated. `bunx tsc --noEmit` clean.

### DIFF
--- a/apps/web/lib/__tests__/truncate-diff.test.ts
+++ b/apps/web/lib/__tests__/truncate-diff.test.ts
@@ -1,0 +1,24 @@
+import { describe, it, expect, mock } from "bun:test";
+
+// github.ts pulls in server-only (via github-app-config) — stub before import.
+mock.module("server-only", () => ({}));
+mock.module("@octopus/db", () => ({ prisma: {} }));
+
+const { truncateDiff } = await import("@/lib/github");
+
+describe("truncateDiff", () => {
+  it("returns small diffs unchanged", () => {
+    const diff = "diff --git a/x b/x\n+hello\n";
+    expect(truncateDiff(diff)).toBe(diff);
+  });
+
+  it("truncates oversized diffs and marks them (never silent)", () => {
+    const big = "x".repeat(400_000);
+    const out = truncateDiff(big);
+    expect(out.length).toBeLessThan(big.length);
+    expect(out).toContain("[... diff truncated");
+    // The default cap (300k) is well above the old 30k, so a 300k diff is NOT
+    // truncated — everyday PRs now review in full.
+    expect(truncateDiff("y".repeat(300_000)).includes("[... diff truncated")).toBe(false);
+  });
+});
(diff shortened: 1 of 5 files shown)


cemoso:
Good catch — replaced the marker sniffing with a structured flag: `getPullRequestDiffViaFiles` now returns `{ diff, truncated }` and the caller uses the boolean for the LargePrError handoff, so a PR whose own content contains the marker string can't false-positive. The marker stays in the diff text for the model, just not used for detection. Thanks @octopus.

cemoso:
Fixed: `getPullRequestDiffViaFiles` now tracks an explicit `hitCap` flag (set when the size cap breaks the loop) and returns that as `truncated`, instead of inferring from the final joined length — the exact-boundary case can't misread as complete. Thanks @octopus.

cemoso:
Addressed all three:
- **Consistency (real catch):** GitLab **and** Bitbucket both still hardcoded the silent 30k cap, so large MRs/PRs there were truncated + scored on a partial view too. Extracted \`lib/diff-truncate.ts\` (env-tunable \`MAX_DIFF_CHARS\`, default 300k) and wired it into **all three** providers — one shared limit, no drift.
- **Code Quality / Error Handling (marker edge):** the \`/files\` reconstruction now appends the truncation notice whenever it hit the cap (covers the exact-boundary case the length check missed).

Test now imports from the dependency-free shared module. Ready for re-review, @octopus.

PR #712, opened 2026-07-30
feat(review): exclude generated files from review before the size cap (#1429)

cemoso:
## Problem (follow-on to the truncation fix, found by another reviewer)

Even after raising the review cap to 300k, a single large **generated** file defeats it. In the reported PR, Drizzle's snapshot JSON is **12,970 lines (~500k+ chars)** — it alone exceeds the cap, sorts **early** alphabetically (`packages/db/drizzle/meta/`), gets truncated in first, and the real hand-written file after it (`packages/shared/src/constants.ts` — which held the "is this account a bank" function everything depends on) was **never reviewed**. It recurs on **every migration PR**. And filtering ran *after* the fetch-time truncation, so it couldn't help.

## Fix — filter generated/ignored files BEFORE the cap

- **`lib/generated-files.ts`**: curated `DEFAULT_GENERATED_PATTERNS` (lockfiles, `**/drizzle/meta/**`, `*.snap`, `*.min.*`, `*.map`, protobuf output…) **plus** parsing the repo's **`.gitattributes` `linguist-generated`/`-diff`** markers. `splitDiffByIgnore` returns the kept diff + the skipped file list.
- **Separate the raw-fetch ceiling from the review cap**: `MAX_FETCH_DIFF_CHARS` (default 1.5M, env-tunable) for the provider fetch, `MAX_DIFF_CHARS` (300k) for review. Fetches (GitHub/GitLab/Bitbucket) now return whole files up to the ceiling; the reviewer **excludes generated + `.octopusignore` files, then caps** — so generated files never consume the budget.
- **`reviewer.ts`** fetches `.gitattributes`, excludes generated + `.octopusignore` before the cap, and adds a footnote listing skipped generated files. **Community path** applies the defaults + cap too.
- Conservative defaults (no `**/generated/**` catch-all) to avoid hiding real code; repos extend via `.gitattributes`.

## Result
Hand-written files in migration/lockfile PRs now get reviewed instead of being dropped for generator output — closes the structural blind spot. The reporter's `.gitattributes linguist-generated` experiment now "just works."

## Tests
`splitDiffByIgnore` excludes generated (drizzle/meta, lockfiles, snapshots) while keeping real files; `.gitattributes` parsing honors `linguist-generated`/`-diff` and `=false`. `bunx tsc --noEmit` clean.

### DIFF
--- a/apps/web/lib/__tests__/generated-files.test.ts
+++ b/apps/web/lib/__tests__/generated-files.test.ts
@@ -1,0 +1,49 @@
+import { describe, it, expect } from "bun:test";
+import {
+  buildGeneratedMatcher,
+  splitDiffByIgnore,
+  parseGitattributesGenerated,
+} from "@/lib/generated-files";
+
+function section(path: string): string {
+  return `diff --git a/${path} b/${path}\n--- a/${path}\n+++ b/${path}\n@@ -1 +1 @@\n-x\n+y\n`;
+}
+
+describe("splitDiffByIgnore (default generated patterns)", () => {
+  it("excludes generated files but keeps hand-written ones", () => {
+    const diff =
+      section("packages/db/drizzle/meta/0007_snapshot.json") +
+      section("bun.lock") +
+      section("src/__snapshots__/foo.test.ts.snap") +
+      section("packages/shared/src/constants.ts");
+    const { kept, skipped } = splitDiffByIgnore(diff, buildGeneratedMatcher());
+
+    expect(skipped).toContain("packages/db/drizzle/meta/0007_snapshot.json");
+    expect(skipped).toContain("bun.lock");
+    expect(skipped).toContain("src/__snapshots__/foo.test.ts.snap");
+    expect(kept).toContain("packages/shared/src/constants.ts");
+    expect(kept).not.toContain("drizzle/meta");
+  });
+});
+
+describe("parseGitattributesGenerated", () => {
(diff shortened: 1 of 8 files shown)


cemoso:
Addressed both:
- **Internal-cli threshold (intent confirmed + documented):** yes, intentional. The clone-based handoff now triggers at the raw-fetch ceiling (`MAX_FETCH_DIFF_CHARS`), not the review cap — below it the standard path fetches the whole diff, excludes generated/ignored files, then caps, which reviews most large PRs *fully* (better than the summary-only large-PR path). internal-cli is reserved for genuinely enormous raw diffs. Added a comment saying so.
- **Truncation marker edge:** if the raw diff hit the fetch ceiling but section-filtering dropped the marker (last raw file excluded), the reviewer now re-appends the notice, gated by a **length check** (`rawDiff.length >= MAX_FETCH_DIFF_CHARS`), not marker sniffing — so the truncation signal is never silently lost.

cemoso:
Fixed the negation: `.gitattributes` `linguist-generated=false` now emits an `ignore` `!pattern` (order-preserved), so a repo can re-include a file that a broader rule or a built-in default (e.g. `*.snap`) would otherwise exclude. Added tests for the negation + default-override. On the consistency note: verified all *live* provider-diff consumers re-cap after the raised fetch ceiling — `reviewer.ts` and `community-review.ts` both filter+`truncateDiff`; `provider.ts`'s method has no callers. Ready for re-review, @octopus.

PR #721, opened 2026-08-01
ci(release): gate releases on a customer-facing changelog entry + stamp it in bump-version

cemoso:
## Why
The public `/docs/changelog` (rendered from `CHANGELOG.md`) had drifted to 1.0.62 while we shipped through 1.0.92. This makes keeping it current part of the release step — **without** auto-generating customer-facing prose from commit messages (those are dev-facing).

## What
- **`release.yml` — `verify-changelog` gate**: a new job that the build jobs (`selfhost`, `prod`) `needs:`, so it gates the whole release. It fails a `v*` tag build if `CHANGELOG.md` has no `## [<version>]` section, with a message pointing at `bump-version.sh`. Enforcement only — never writes prose.
- **`bump-version.sh` — changelog stamp**: after bumping `package.json` versions, it promotes the accumulated `## [Unreleased]` notes into a dated `## [X.Y.Z]` section and opens a fresh `[Unreleased]` (Keep a Changelog convention). Human-written notes are just moved into place; if `[Unreleased]` is empty it warns and inserts a placeholder. Idempotent (skips if the version section already exists).

## Flow going forward
Add customer-facing bullets under `[Unreleased]` as work lands → `scripts/bump-version.sh X.Y.Z` → review `CHANGELOG.md` → commit → tag. A tag without an entry now fails fast instead of shipping a stale changelog.

Verified the stamp transform on a sample changelog (Unreleased → dated section, older entries intact) and `bash -n` on the script.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,8 +29,31 @@ permissions:
   packages: write
 
 jobs:
+  # ── changelog gate ─────────────────────────────────────────────────────────
+  # Block a release that has no customer-facing CHANGELOG.md entry for its
+  # version, so the public /docs/changelog can never silently fall behind.
+  # Content stays human-curated (scripts/bump-version.sh stamps [Unreleased]
+  # into the version) — we only ENFORCE the entry exists, never auto-write prose.
+  verify-changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - name: Require a CHANGELOG.md entry for this version
+        run: |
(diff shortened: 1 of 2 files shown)


cemoso:
All three addressed:
- **Code Quality (real bug):** switched to a replacer **function** in `bump-version.sh` so `$` in notes isn't treated as a `String.replace` pattern. Verified our own "$5 / $25 per million tokens" line (and even `$&`) round-trips intact.
- **Security:** `verify-changelog` now declares `permissions: contents: read` instead of inheriting `contents/packages: write`.
- **Error Handling:** the gate now emits a clear `::error::` when `CHANGELOG.md` is missing, instead of a raw grep failure.

PR #747, opened 2026-08-13
feat(admin): merge-orgs endpoint to consolidate one org into another

cemoso:
## What

A vendor-admin machine endpoint `POST /api/admin/merge-orgs` (guarded by the shared `ADMIN_API_SECRET` bearer, inert on self-host) that consolidates one organization into another, plus the core logic in `lib/merge-orgs.ts`.

Driven over loopback by octopus-deploy's `merge-orgs-wdc.yml` (backup → dry-run → apply), mirroring `seed-docs-wdc`.

## Behavior

- **Single Postgres interactive transaction** re-points every org-scoped table from source → target. `dryRun` (the default) executes every statement then **rolls back**, returning exact affected-row counts — a zero-risk preview against live data. Callers must pass `dryRun:false` to commit.
- **Balances** summed onto the target; **survivor renamed** (optional); **source API tokens revoked** by default (so they can't silently reach the target's repos); **source org soft-deleted** (`deletedAt`) — never hard-deleted, so no destructive FK cascade (matches `deleteOrganization` in `actions.ts`).
- **Compound-unique tables** (`organization_members`, `day_summaries`, `incident_comms`, `coupon_redemptions`, `user_presences`) drop colliding source rows before moving; members demote a moved `owner` to `admin` and keep the target's own owner.
- **Qdrant**: on apply (after commit) rewrites the `orgId` payload across `knowledge_chunks`, `review_chunks`, `chat_chunks`, `flowchart_chunks`, `feedback_patterns` (code_chunks is `repoId`-keyed, untouched) — otherwise those vectors silently orphan.

## Safety

- Table classification verified against `packages/db/prisma/schema.prisma` (every `organizationId`-bearing table enumerated; compound uniques handled).
- Preconditions: source/target exist, neither already deleted, new slug not taken.
- No schema change. `tsc --noEmit` clean.
- The real acceptance check is the production **dry-run** (rollback) row counts, run before any apply.

### DIFF
--- a/apps/web/app/api/admin/merge-orgs/route.ts
+++ b/apps/web/app/api/admin/merge-orgs/route.ts
@@ -1,0 +1,52 @@
+import { NextRequest, NextResponse } from "next/server";
+import { isAdminApiAuthorized } from "@/lib/admin-auth";
+import { mergeOrgs } from "@/lib/merge-orgs";
+
+// Vendor admin machine endpoint: consolidate one organization into another.
+// Guarded by the shared ADMIN_API_SECRET bearer (inert on self-host). Driven by
+// octopus-deploy's merge-orgs-wdc.yml over loopback, mirroring seed-docs.
+//
+// dryRun defaults to TRUE: the caller MUST pass dryRun:false explicitly to
+// commit, so an accidental call can never mutate data.
+export const runtime = "nodejs";
+export const dynamic = "force-dynamic";
+
+export async function POST(request: NextRequest) {
+  if (!isAdminApiAuthorized(request)) {
+    return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+  }
+
+  let body: Record<string, unknown>;
+  try {
+    body = (await request.json()) as Record<string, unknown>;
+  } catch {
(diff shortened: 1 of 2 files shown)


cemoso:
@octopus addressed both scored concerns in the latest commit:

**Membership dedupe (Code Quality):** the dedupe now works on active rows. Because the `[organizationId, userId]` unique slot survives a soft delete, the old `NOT IN (all target rows)` would have dropped an active source member who had a *soft-deleted* row in the target. New flow: (1) reactivate a soft-deleted target row when the user is active in the source, (2) move only active source rows with no target row (demoting a moved owner to admin), (3) soft-close leftovers.

**Soft-delete consistency:** the four hard-`DELETE` tables (`day_summaries`, `user_presences`, `incident_comms`, `coupon_redemptions`) have **no `deletedAt` column**, so soft-delete isn't possible there — only colliding *source* rows are dropped and each is regenerable/ephemeral/a kept-on-target duplicate. Added a comment making this deliberate distinction explicit vs. the soft-deletes used for the org/members/tokens.

`tsc --noEmit` clean. Please re-review.

PR #762, opened 2026-08-20
feat(onboarding): connect-first + GitHub reconnect prompts + funnel tracking

cemoso:
## Why

Of 326 orgs that registered but never ran a review, 269 (82%) never connected a GitHub App or any repo. They land on the analytics dashboard after signup, see an all-zero grid whose only connect prompt is a dismissible banner, and bounce. This PR makes connecting a repo the obvious first action and instruments the funnel so the drop-off becomes measurable. Provider-agnostic on purpose — no redirect that assumes GitHub.

## What

1. **Connect-first dashboard state** — `apps/web/app/(app)/dashboard/page.tsx`: when the org has no provider connected AND zero repos, render a prominent, non-dismissible "Connect your code to start reviewing" hero (GitHub / GitLab / Bitbucket, with the existing connect flows) instead of the empty analytics grid. Implemented as a `hero` prop on the existing `ProvidersBanner`, so the provider cards, Bitbucket workspace dialog, and GitLab OAuth dialog are reused verbatim. The early return also skips the chart queries, which are all empty for this cohort. Orgs with any repo or provider see the dashboard unchanged.
2. **Repositories empty state** — `repositories-content.tsx`: "No repositories yet" now has a primary **Connect GitHub** button (same `/api/github/install?returnTo=/repositories` action as the toolbar's Manage GitHub) plus a link to connect GitLab or Bitbucket via Settings → Integrations. Shown only when not filtering by search.
3. **Funnel instrumentation (purely additive)** —
   - `integration.install_started` audit log in `/api/github/install` (actor, org, returnTo)
   - `integration.connected` audit log on successful bind in `/api/github/callback`
   - `integration.install_failed` audit log with the reason on every error redirect in `/api/github/callback` (single hook in `errorRedirect`, so all ~20 failure reasons are captured)
   - `trackEvent("cta_click", ...)` on the hero and ProvidersBanner Connect GitHub clicks, distinguished by `location` (`dashboard_connect_hero` vs `dashboard_providers_banner`), following the settings/integrations pattern

Deliberately unchanged: the post-signup redirect in `complete-profile/actions.ts` — the provider-agnostic hero handles that cohort without assuming GitHub.

## Verification

`bunx tsc --noEmit` in `apps/web`: clean.

### DIFF
--- a/apps/web/app/(app)/dashboard/page.tsx
+++ b/apps/web/app/(app)/dashboard/page.tsx
@@ -202,6 +202,31 @@ export default async function DashboardPage({
   const gitlabConnected = !!gitlabIntegration;
   const bannerDismissed = cookieStore.get("providers_banner_dismissed")?.value === "1";
 
+  // Connect-first empty state: an org with no provider and no repos gets one
+  // clear action instead of an all-zero analytics grid. Not dismissible —
+  // there is nothing else to show until code is connected. Also skips the
+  // chart queries below, which would all be empty for this cohort.
+  if (!githubConnected && !bitbucketConnected && !gitlabConnected && totalRepos === 0) {
+    return (
+      <div className="mx-auto max-w-6xl p-6 md:p-10">
+        <div>
+          <h1 className="text-2xl font-semibold tracking-tight">Dashboard</h1>
+          <p className="text-muted-foreground mt-1 text-sm">
+            Overview of your repositories and integrations.
+          </p>
+        </div>
+        <ProvidersBanner
+          hero
+          githubConnected={githubConnected}
+          bitbucketConnected={bitbucketConnected}
+          gitlabConnected={gitlabConnected}
+          githubAppSlug={githubAppSlug}
+          gitlabRedirectUri={process.env.GITLAB_REDIRECT_URI ?? null}
+        />
+      </div>
+    );
+  }
+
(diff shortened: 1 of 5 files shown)


cemoso:
Added a second commit (`ad69618`): a **GitHub reconnect prompt**.

Prod scan found ~9 non-community orgs whose GitHub App was uninstalled (`githubInstallationId` nulled) but that still have active indexed GitHub repos — so every PR webhook is silently dropped and they get zero reviews while the product still looks connected. Some were previously active (Open Software 79 PRs, ian 15, Kaleb 29).

The dashboard now detects `!githubConnected && repos.some(github)` and shows a clear "GitHub connection removed — reconnect" banner linking to `/api/github/install`, and suppresses the generic providers banner in that case to avoid double CTAs. Detection only — no webhook or repo-state changes.

cemoso:
@octopus Fix applied for the 🔵 finding — ready.

## Changes Made
| # | Finding | Action |
|---|---------|--------|
| 1 | Install-start telemetry can block redirect on transient audit failure (`install/route.ts:L52`) | ✅ Fixed in `281b22c` — `writeAuditLog` is now fire-and-forget (`void … .catch(() => {})`), so telemetry is off the onboarding critical path |

PR was already 4/5 (gate met); this addresses the sole info-level finding.

